PR #1, opened 2026-04-21
Feat/mvp

arm64be:


zebster-cmd:
# Code review: `feat/mvp`

6 commits, ~700 lines of real code plus ~5.4k lines of JSON fixtures (not reviewed).

## Blockers
None strictly blocking, but two issues are worth addressing before merge.

## High-value fixes

**1. Event listeners leak on `destroy()` — `theia-panel/src/index.ts:83, 109, 116, 134, 151`**
Five `element.addEventListener` calls (mousemove×2, mousedown, mouseup, wheel), but `destroy()` (line 180) disposes scene/nodes/picker/etc. and never calls `removeEventListener`. If a consumer calls `mount()` → `destroy()` → `mount()` on the same element, listeners stack and every event fires twice. Fix: store handler refs, remove them in `destroy()`.

**2. `started_at` silently falls back to ingest time — `theia-core/theia_core/ingest.py:164`**
```python
started_at = min(timestamps) if timestamps else datetime.now(UTC)
```
If neither `meta.timestamp` nor any message has a timestamp, the session gets stamped with "now", which poisons downstream time-series analysis and makes the session look brand-new every ingest. Better: raise / warn, or fall back to epoch so it's obviously sentinel.

## Suggestions

**3. `salience` always 0.5 for JSONL path — `ingest.py`**
Fixture JSON sessions carry `salience: 0.9`; JSONL-derived `MemoryEvent`s default to 0.5 because `_infer_memory_events` doesn't pull it from tool args. Possibly intentional (real Hermes logs don't emit salience), but worth a comment.

**4. Pan math assumes rect aspect == camera aspect — `index.ts:125–128`**
`panX = -(dx / rect.width) * worldW` is only 1:1 accurate when rect aspect matches `(right-left)/(top-bottom)`. The `resize` handler keeps them in sync, but if the camera frustum is ever letterboxed this drifts. Minor.

**5. Duplicate `picker.onHover` registration — `index.ts:87 and 175`**
Two separate handlers registered. Works, but consolidate into one to avoid surprises.

**6. `warnings` imported inside loop — `ingest.py:201`** — nit, move to top.

**7. CI workflows have no `concurrency` key — `.github/workflows/*.yml`**
Rapid pushes to `main` spin up redundant runners. Add:
```yaml
concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: true
```

**8. CI workflows use default `GITHUB_TOKEN` permissions**
Add `permissions: { contents: read }` (plus `pull-requests: write` on `preview.yml` if it needs to comment). Defense-in-depth, cheap.

**9. `cross_search.py` defensive `source_session_id is None` check — `cross_search.py:15`**
The type is non-`Optional` and the ingest filter (`ingest.py:99-100`) already drops empty values. Fix works but signals a loose type contract — consider marking the field `Optional[str]` or trusting the upstream filter.

## Nice work
- Commit `0bd7dd7` (in-place edge updates): confirmed the BufferGeometry is reused and `needsUpdate` signaled correctly (`Edges.ts:139`). Good perf win.
- New JSONL test `test_ingest_jsonl.py` (77 lines) looks reasonable for the new path.
- Workflows cache pip + npm — good.

## Summary
Two real things to fix pre-merge: listener leak in `mount/destroy` cycle and the `now()` fallback for missing timestamps. The rest is polish.

zebster-cmd:
# Code review follow-up: `feat/mvp` (re-review)

The review above was on an outdated local snapshot. Re-reviewed against current `origin/feat/mvp` HEAD (`7b849af`): **41 commits, +4,265 / -244 lines** (excluding ~2.7k lines of `examples/*.json` fixtures).

The five findings below (listener leak, `datetime.now()` fallback, salience default, duplicate `onHover`, pan math, `warnings` import, `cross_search.py` guard, CI hygiene) from the original review still stand — line numbers have shifted, updated inline below. The new findings are in the Post-processing (`Post.ts`), physics (`Simulation.ts`), and detect (`tool_overlap.py`, `memory_share.py`) code added after the original review.

## Blockers
None strictly blocking — but some real leaks and a quadratic hot path worth addressing before merge.

## New high-value fixes (post-original-review commits)

**A. `ditherTarget`, dither/overlay materials, and quads leak on `destroy()` — `theia-panel/src/scene/Post.ts:164, 214–240, 299`**
`createPost` creates `ditherTarget`, `ditherMaterial`, `overlayMaterial`, `ditherQuad`, and `overlayQuad`, but none are exposed on `PostContext`. `index.ts` `destroy()` (lines 192–195) disposes only `edgesTarget`, `preBloomTarget`, `sceneTarget`, and `composer`. Every `mount → destroy` cycle leaks a render target plus two `ShaderMaterial`s and two `FullScreenQuad`s (GPU memory + compiled shader programs). Expose them on `PostContext` (or add a `post.dispose()`) and call it in `destroy()`.

**B. `ditherTarget` is never resized — `Post.ts:164, 288–297`**
`resize()` handles `edgesTarget`, `preBloomTarget`, `sceneTarget`, `composer`, and `bloom` — but not `ditherTarget`. After any window resize, the dither pass samples a stale-size texture and the output stretches/clips. Add `ditherTarget.setSize(w2 * dpr2, h2 * dpr2)`.

**C. `forceCluster` is O(N² · avg_degree) per tick — `theia-panel/src/physics/Simulation.ts:47–83`**
On every alpha step the force:
- Rebuilds the adjacency map from scratch (lines 50–55).
- For each neighbor id, calls `nodes.find(x => x.id === nid)` (line 64) — O(N) linear scan inside the hot loop.

Net: O(N² · avg_degree) per tick, running while the simulation is warm. Adjacency + an `id → PhysicsNode` map should be built once in `force.initialize` and cached. Noticeable today, a frame-budget killer once the graph grows.

## New suggestions

**D. Dead-code `or False` in skill_view branch — `theia-core/theia_core/detect/tool_overlap.py:64`**
```python
if last_managed or False:  # current is view, so need last to have managed
```
`or False` is a no-op; the condition reduces to `if last_managed`. Looks like a refactor leftover — drop it (the comment also drifts from the code).

**E. Unused `defaultdict` imports — `tool_overlap.py:4`, `memory_share.py:3`**
Neither module uses `defaultdict`. Remove.

**F. `NODE_GLOW_TEXTURE` built at module import — `theia-panel/src/scene/Nodes.ts:5, 7–28`**
Top-level `makeGlowTexture()` touches `document`, so importing `Nodes.ts` throws in any non-DOM context (SSR, node-based tests). Lazy-init inside `createNodes` (and then share via a module-scoped cache after first call).

**G. `Nodes.setPosition` decomposes/recomposes per node per frame — `Nodes.ts:111–117`**
`getMatrixAt → decompose → position.set → updateMatrix → setMatrixAt` runs N times per tick (called from `index.ts` lines 63–67). Since `nodeSizes[i]` is already known, compose the matrix directly from `(x, y, sz)` and skip the decompose/updateMatrix round-trip.

## Original findings — line numbers updated

- **Event listener leak on `destroy()`** — now `index.ts:90, 116, 123, 141, 158`; `destroy()` at line 187. Same fix.
- **`started_at` fallback to `datetime.now()`** — now in **both** parse paths: `ingest.py:203` (session-archive) and `ingest.py:245` (JSONL).
- **`salience` default 0.5 on inferred memory events** — still `ingest.py:97` in `_infer_memory_events`.
- **Duplicate `picker.onHover`** — now `index.ts:94 and 182`.
- **Pan math aspect assumption** — now `index.ts:131–136`.
- **`warnings` imported inside loop** — now `ingest.py:310`.
- **`cross_search.py:15`** defensive `source_session_id` guard — unchanged.
- **CI `concurrency` / `permissions`** — still missing.

## Nice work (new)
- `examples/hermes_graph.json` generated from real Hermes JSONL — good end-to-end signal.
- `_parse_jsonl` / `_parse_session_json` / `_parse_fixture_json` split is cleaner than a single mega-parser, and the `_is_*_format` sniffers are explicit about what each accepts.
- Shared `ShaderMaterial`s cached across `Edges.rebuild` calls — avoids shader recompiles on filter toggle.
- Post chain (render → composite → pre-bloom save → bloom → scene save → dither → overlay) is well-separated and readable despite its length.

## Updated summary
The big new issues all live in the post-processing and physics code added after the original review:
- **`Post.ts`**: `ditherTarget` isn’t resized and the dither/overlay resources aren’t exposed for disposal — both are real leaks.
- **`Simulation.ts`**: `forceCluster` is quadratic in node count on every tick.

Most of the original findings still stand. Nothing blocks merge outright, but items A–C plus the original #1 and #2 are the ones worth fixing before shipping.

PR #4, opened 2026-04-23
feat(panel): node summaries, search bar, richer edges + review fixes

arm64be:
- add summary/initial_prompt extraction to emitter and schema

- add SearchBar with Map-based lookup and mousedown-blur fix

- add animated focusOn to Scene with RAF cancellation

- refactor SidePanel edge rendering with per-kind details and colors

- add summary/initial_prompt display to Tooltip and SidePanel

- DRY escape/truncate into ui/utils.ts

- defensive isinstance(msg, dict) guard in emit.py

### DIFF
--- a/examples/graph.json
+++ b/examples/graph.json
@@ -1,9 +1,9 @@
 {
   "meta": {
-    "generated_at": "2026-04-21T15:12:16.314028Z",
+    "generated_at": "2026-04-22T22:07:00.633082Z",
     "source_count": 3,
-    "projection": "pca",
-    "feature_dim": 4
+    "projection": "umap",
+    "feature_dim": 5
   },
   "nodes": [
     {
@@ -11,12 +11,14 @@
       "title": "Alpha \u2014 writes auth memory",
       "started_at": "2026-04-18T09:00:00Z",
       "duration_sec": 1000.0,
-      "tool_count": 3,
+      "tool_count": 4,
       "message_count": 50,
       "model": "claude-opus-4-7",
+      "summary": "Designed the authentication middleware architecture and stored it in persistent memory. Covered JWT token validation, refresh flows, and middleware integration patterns.",
+      "initial_prompt": null,
       "position": {
-        "x": 5.263409943150801e-16,
-        "y": 0.5029945449101046
+        "x": 0.19980697827081964,
+        "y": 0.40134980674379
       },
       "features": null
     },
@@ -28,9 +30,11 @@
       "tool_count": 2,
       "message_count": 32,
       "model": "claude-opus-4-7",
+      "summary": null,
+      "initial_prompt": "I need to implement auth middleware for our API. Can you help me understand the best approach for JWT validation and refresh token handling?",
       "position": {
-        "x": -0.9678580071207084,
-        "y": -0.25149727245505216
+        "x": -0.9910909253370497,
+        "y": -0.13318700279888648
       },
       "features": null
     },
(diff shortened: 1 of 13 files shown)


zebster-cmd (commented):
## Code Review — PR #4

### Overview

This PR adds node summaries/initial prompts, a search bar, animated camera focus, and richer edge rendering to the panel. It also extracts shared `escape`/`truncate` helpers and adds defensive guards in the Python emitter. Overall a solid, well-scoped feature set. A few items to address below.

---

### Bugs / Correctness

1. **`focusOn` RAF leak on dispose** — `Scene.ts:79-100`: The `focusOn` animation stores a `rafId` and cancels on re-entry, but `dispose()` (`Scene.ts:110-114`) does **not** cancel an in-flight animation frame. If `destroy()` is called while a focus animation is running, the RAF callback fires after disposal, calling `apply()` on a detached camera/renderer. Add `if (rafId !== null) cancelAnimationFrame(rafId);` to `dispose()`.

2. **`escape()` does not handle single-quote (`'`)** — `utils.ts:1-5`: The function escapes `& < > "` but not `'`. Since the SidePanel and SearchBar inject escaped values into HTML attribute contexts (e.g., `data-index`), a node title containing `'` would not be exploitable in the current code (since `data-index` is always a number), but the summary/initial_prompt text is injected into innerHTML where event handlers could theoretically be added. Low risk currently, but adding `"'": "&#39;"` to the replacement map would close the gap.

3. **Stale `aspect` captured at init** — `Scene.ts:19-45`: The `apply()` function uses the initial `aspect` from line 19 forever, while `resize()` recalculates it locally. This means `focusOn` and `pan` (which call `apply()`) will use the **original** aspect ratio, not the current one after a resize. Consider making `aspect` a `let` that `resize()` updates, or have `apply()` read it dynamically.

---

### Design / Architecture

4. **`resultByIndex` map is redundant** — `SearchBar.ts:77-84`: `results` is built as an array, and then a parallel `Map<number, SearchResult>` is created keyed by `index`. Since `index` is the node array index, you could just look up `results.find(r => r.index === idx)` or, better, use a single `Map<number, SearchResult>` and derive the array from `map.values()` to avoid maintaining two parallel structures.

5. **`renderEdge` evidence casting is repetitive** — `SidePanel.ts:68-117`: Each edge kind branch casts `evidence as Record<string, unknown>` repeatedly. Consider extracting `const ev = (e.evidence ?? {}) as Record<string, unknown>;` once at the top of `renderEdge` and reusing it.

6. **Edge filter now drops `tool-overlap`** — `examples/graph.json` lost the `tool-overlap` edge between alpha-beta and beta-gamma. Combined with `DEFAULT_KINDS` in `index.ts:25-28` only including `memory-share` and `cross-search`, the example data now has only 1 tool-overlap edge that is hidden by default. This seems intentional but worth confirming — the `renderEdge` function has careful styling for tool-overlap evidence, but no way to see it in the demo without toggling the filter.

---

### Nits / Style

7. **Long HTML template strings** — `SidePanel.ts:24-28` and `Tooltip.ts:17-21`: These single-expression ternary chains building HTML are hard to read. Consider extracting a small helper like `renderSummaryBlock(node)` to keep the `show()` method scannable.

8. **Magic number `280`/`180` for truncation** — `SidePanel.ts:27` truncates at 280 chars, `Tooltip.ts:18,20` at 180 chars. These could be named constants (e.g., `const TOOLTIP_MAX_CHARS = 180`) for clarity and easy tuning.

9. **`searchBar.dispose` missing from main branch `destroy()`** — Actually, this is handled correctly in the PR at `index.ts:215`. Good.

10. **`valid_edges` duplication** — `emit.py:73-74`: This logic already existed in `main` (lines 45-46). The PR diff shows it being added, but comparing the full files confirms both branches have it. This is correct; no issue.

---

### Schema

11. **`summary` and `initial_prompt` allow `null`** — `graph.schema.json:34-35` uses `"type": ["string", "null"]` which is correct and matches the Python emitter returning `None`. The TypeScript type uses `summary?: string` (optional, not nullable). Technically the schema allows an explicit `null` value, while TS treats it as optional-string-or-undefined. This works in practice because JSON `null` coerces appropriately when read, but for strict correctness the TS type could be `summary?: string | null`.

---

### Summary

The PR is well-structured and the features are cohesive. The **stale aspect ratio in `apply()`** (item 3) and the **RAF leak on dispose** (item 1) are the most impactful issues — both can cause subtle rendering bugs. The rest are minor improvements. Approving with the expectation that items 1 and 3 will be addressed.

PR #5, opened 2026-04-23
feat: 3D star cluster transition + review fixes

arm64be:


zebster-cmd (commented):
## Code Review: feat: 3D star cluster transition + review fixes

### Build Status: FAILS

The branch does not compile cleanly. There are **7 TypeScript errors** that must be fixed before merge.

---

### Blocking Issues

#### 1. TS errors in `Scene.ts` — array index access on `matrixWorld.elements`

`theia-panel/src/scene/Scene.ts:64-66`

```ts
target.x += (-m[0] * dxPixel + m[4] * dyPixel) * panSpeed;
target.y += (-m[1] * dxPixel + m[5] * dyPixel) * panSpeed;
target.z += (-m[2] * dxPixel + m[6] * dyPixel) * panSpeed;
```

`camera.matrixWorld.elements` is typed as `number[]`, so `m[0]`, `m[4]`, etc. are `number | undefined`. Either use non-null assertions (`m[0]!`) or destructure with defaults. Consider extracting the right/up vectors using `THREE.Vector3.setFromMatrixColumn()` for both correctness and readability.

#### 2. TS error in `index.ts:81` — possibly undefined uniform access

```ts
nodes.material.uniforms.uCameraPos.value.copy(ctx.camera.position);
```

`ShaderMaterial.uniforms` is `Record<string, IUniform | undefined>`, so `.uCameraPos` might be `undefined`. Add a non-null assertion or a guard. Better yet, see the suggestion below about not exposing raw material on the interface.

---

### Non-blocking Issues

#### 3. Leaking `material` on the `NodeLayer` interface

`theia-panel/src/scene/Nodes.ts:108`

Exposing `material: THREE.ShaderMaterial` on the public interface just so `index.ts` can poke at a uniform is a code smell. A cleaner approach:

```ts
// in NodeLayer interface
setCameraPosition(pos: THREE.Vector3): void;

// in createNodes implementation
setCameraPosition(pos) {
  material.uniforms.uCameraPos!.value.copy(pos);
},
```

This keeps the shader implementation detail private and also eliminates the TS error in `index.ts`.

#### 4. Raycasting vs. billboard shader mismatch

`theia-panel/src/scene/Picker.ts` uses `raycaster.intersectObject` against the InstancedMesh, but the new vertex shader applies a billboard transform in **view-space** (shifting `position.xy` onto the camera-facing plane). The raycaster has no knowledge of this custom shader — it tests against the original geometry in world-space. This means pick accuracy will degrade as the camera rotates away from default orientation, since the visual quad and the raycasted quad diverge. At extreme angles, hovers may miss or hit the wrong node.

Consider switching the pick logic to screen-space distance checks (project instance centers to NDC, find nearest to cursor) or updating the bounding sphere per-instance to match the billboard size.

#### 5. Duplicated hash functions

`theia-panel/src/scene/Nodes.ts:31-38` (`hashString`) and `theia-panel/src/physics/Simulation.ts:128-135` (`hashId`) have identical FNV-1a bodies but map to different ranges (`[0,1)` vs `[-1,1]`). Extract the core hash to a shared utility and parameterize the output range.

#### 6. O(n*m) neighbor lookup in `forceCluster` on every tick

`theia-panel/src/physics/Simulation.ts:69`

```ts
const neighbor = nodes.find((x) => x.id === nid);
```

This `Array.find` runs inside a hot loop (every tick, for every neighbor of every node). With a large graph, this is quadratic. Build a `Map<string, PhysicsNode>` at the start of the force function:

```ts
const nodeMap = new Map(nodes.map(n => [n.id, n]));
// ...
const neighbor = nodeMap.get(nid);
```

#### 7. Edge initial z-positions hardcoded to 0

`theia-panel/src/scene/Edges.ts:88,91` — during `rebuild()`, edge endpoint z-coordinates are set to `0`. This is overwritten on the first simulation tick via `updatePositions()`, but there may be a single-frame flash where edges render flat before snapping to 3D positions. Consider initializing from the simulation node positions or skipping the first render until a tick completes.

#### 8. Hardcoded dev server port

`theia-panel/vite.config.ts:5` — `port: 5179` is unexplained. If there is a reason (avoiding conflicts with another service), a comment would help. Otherwise, letting Vite pick the default is less surprising.

---

### Summary

The 2D-to-3D upgrade is well-structured overall — the physics forces, camera orbit math, and billboard shader are solid. The main blocker is the 7 TypeScript errors preventing a clean build. After fixing those, the raycaster/billboard mismatch and the `forceCluster` performance issue are the most impactful items to address.

zebster-cmd (commented):
Inline comments on the 7 failing TypeScript checks.

theia-panel/src/scene/Scene.ts
```diff
@@ -68,14 +58,25 @@ export function createScene(container: HTMLElement): SceneContext {
...
     renderer,
     container,
-    pan(dxWorld, dyWorld) {
-      panX += dxWorld;
-      panY += dyWorld;
-      apply();
+    pan(dxPixel, dyPixel) {
+      const panSpeed = radius * 0.0015;
+      const m = camera.matrixWorld.elements;
+      target.x += (-m[0] * dxPixel + m[4] * dyPixel) * panSpeed;
```
zebster-cmd:
```
error TS2532: Object is possibly 'undefined'.
```
`camera.matrixWorld.elements` is typed as `number[]`, so indexing it returns `number | undefined`. Both `m[0]` and `m[4]` trigger this error on this line.

Quick fix: non-null assert each access (`m[0]!`, `m[4]!`), or destructure the elements array up front.

Cleaner fix: use `THREE.Vector3.setFromMatrixColumn(camera.matrixWorld, 0)` for the right vector and column 1 for the up vector — avoids raw index access entirely.

theia-panel/src/scene/Scene.ts
```diff
@@ -68,14 +58,25 @@ export function createScene(container: HTMLElement): SceneContext {
...
+      target.y += (-m[1] * dxPixel + m[5] * dyPixel) * panSpeed;
```
zebster-cmd:
```
error TS2532: Object is possibly 'undefined'.
```
Same issue — `m[1]` and `m[5]` are `number | undefined`.

theia-panel/src/scene/Scene.ts
```diff
@@ -68,14 +58,25 @@ export function createScene(container: HTMLElement): SceneContext {
...
+      target.z += (-m[2] * dxPixel + m[6] * dyPixel) * panSpeed;
```
zebster-cmd:
```
error TS2532: Object is possibly 'undefined'.
```
Same issue — `m[2]` and `m[6]` are `number | undefined`.

theia-panel/src/index.ts
```diff
@@ -77,6 +78,7 @@ export async function mount(
     const t = performance.now() / 1000;
     nodes.setTime(t);
     edges.setTime(t);
+    nodes.material.uniforms.uCameraPos.value.copy(ctx.camera.position);
```
zebster-cmd:
```
error TS18048: 'nodes.material.uniforms.uCameraPos' is possibly 'undefined'.
```
`ShaderMaterial.uniforms` is typed as `Record<string, IUniform | undefined>`, so `uniforms.uCameraPos` may be `undefined`.

A non-null assertion (`uniforms.uCameraPos!.value`) fixes the type error, but the better fix is to stop exposing `material` on the `NodeLayer` interface altogether — add a `setCameraPosition(pos: Vector3)` method to `NodeLayer` and call it here instead. That eliminates this error and keeps shader internals private.

PR #6, opened 2026-04-23
feat: hot-reload polling, graph lifecycle, watcher plugin

arm64be:


zebster-cmd (commented):
## Code Review — PR #6

### Overview

This PR introduces hot-reload polling for the graph visualization, a full graph lifecycle (tear-down + rebuild on update with position preservation), a Vite plugin that watches the sessions directory and re-runs the Python pipeline, and significant physics simulation improvements (cluster force, collision, per-kind link tuning). It also deletes the committed example graph files and gitignores them. Large refactor of `index.ts` (+247/-75) that restructures the entire mount lifecycle.

---

### Bugs / Correctness

**1. `edges` is shared singleton but stored in every `GraphState`** — `index.ts:65-66,229-232`: A single `EdgeLayer` is created at line 65 and shared across all graph reloads. However, `state.edges` references this same object, and `updateGraph` calls `edges.rebuild(...)` through both the direct `edges` variable and `state.edges`. This works today because they are the same reference, but storing it in `GraphState` implies per-graph ownership. If someone later assumes `state.edges` is scoped to a reload and calls `state.edges.dispose()` in teardown (like `state.nodes.dispose()` at line 259), it would destroy the shared singleton. Consider either: (a) creating a new `EdgeLayer` per reload and disposing the old one, or (b) removing `edges` from the `GraphState` interface to make the shared ownership explicit.

**2. `updateGraph` does not close the SidePanel** — `index.ts:241-298`: When the graph is hot-reloaded, the SidePanel may still be showing a node from the previous graph. The node reference it holds becomes stale (the old `TheiaGraph["nodes"][number]` object). If the user interacts with the panel after a reload (e.g., clicking a connection link), it will reference node IDs that may or may not exist in the new graph. Should call `sidePanel.hide()` at the start of `updateGraph`, or re-show it if the same node ID still exists.

**3. `checkForUpdate` bypasses `loadGraph` validation** — `index.ts:305-323`: `loadGraph(graphUrl)` is called for bootstrap (which internally validates `graph.meta`, `graph.nodes`, `graph.edges` existence via `load.ts:7-13`). But the poll at line 308 uses raw `fetch(graphUrl + "?t=" + Date.now())` and casts the result with `as TheiaGraph` with no validation. If the graph file is malformed during a rebuild (e.g., the watcher fires mid-write), the poll will pass garbage data to `updateGraph`. The poll should either reuse `loadGraph` or duplicate its validation.

**4. Watcher only triggers on NEW files, never on content changes** — `vite-plugin-theia-watcher.ts:82`: Once a session file is added to `knownFiles`, it is permanently ignored. The comment says "Existing files being written to (active .jsonl) are ignored" — but if a session file is replaced (deleted and re-created, or overwritten with new content), the watcher won't detect it. This is fine for append-only `.jsonl` ingestion, but the intent should be documented more explicitly since it's a surprising behavior for a "watcher" plugin.

**5. `forceCluster` rebuilds adjacency map every tick** — `Simulation.ts:49-55`: The adjacency `Map<string, Set<string>>` is reconstructed on every simulation tick. With `alphaTarget(0.015)` keeping the simulation alive indefinitely, this runs ~60 times/second. For the current small graph this is fine, but it's O(N+E) per tick for no reason — the topology doesn't change between ticks. Pre-compute the adjacency map in `forceCluster`'s `initialize` or closure.

---

### Design / Architecture

**6. Plugin always loaded, even in production build** — `vite.config.ts:30-36`: `theiaWatcherPlugin` is unconditionally added to `plugins`. While the plugin only activates in `configureServer` (dev mode), it still imports `child_process`, `fs`, etc. and resolves file paths at config load time. If the `.venv` directory or sessions directory doesn't exist, `readdirSync` in the plugin throws during production `vite build`. Consider gating the plugin:
```ts
plugins: [
  ...(command === "serve" ? [theiaWatcherPlugin({...})] : []),
],
```

**7. `forceCollide` uses O(N²) brute-force** — `Simulation.ts:86-113`: The nested loop collision detection is O(N²). d3-force provides `d3.forceCollide()` with quadtree optimization. For a small graph this is fine, but the custom implementation should be noted as not scaling beyond ~1000 nodes.

**8. `index.html` uses top-level `await` without error handling** — `index.html:21-25`: `await mount(...)` at the top level will silently reject if the graph.json fetch fails (no `.catch()`). The old code had `.catch(console.error)` which was more defensive. Consider wrapping in try/catch or adding `.catch()`.

**9. Hardcoded `~/.hermes/sessions` path** — `vite.config.ts:7`: The sessions directory is hardcoded to `$HOME/.hermes/sessions`. This should probably use an environment variable (e.g., `process.env.THEIA_SESSIONS_DIR`) with the hardcoded path as a fallback, so CI/team members with different layouts can override it.

**10. `graph.json` and `hermes_graph.json` deleted + gitignored** — `.gitignore:18-20` and the -2845 deletion: This removes the example data that other PRs (like #4) are still modifying. Merging this PR will conflict with #4. The gitignore entries make sense for generated output, but there should be a way to bootstrap the dev environment without running the Python pipeline first. Consider either: keeping a `graph.example.json` that is committed, or having `npm run dev` auto-run the pipeline if graph.json is missing.

---

### Nits / Style

**11. `state.edges` aliasing** — Throughout `index.ts`, sometimes `edges` is accessed directly (line 85, 96, 266, 357), sometimes through `state.edges` (line 189). Since they are the same object, this is technically fine but inconsistent. Pick one convention.

**12. `import * as THREE` only used for `new THREE.Scene()`** — `index.ts:1,64`: The star import pulls the entire THREE namespace but only `THREE.Scene` is used in this file. Consider `import { Scene } from "three"` to match the selective import style used elsewhere.

**13. Simulation export types** — `Simulation.ts:10,21`: Exporting `PhysicsNode` and `PhysicsLink` is good for the new lifecycle code, but adding `radius` to `PhysicsNode` and `kind` to `PhysicsLink` changes the interface. These are only used internally by the simulation forces. The radius could live on a separate internal type to keep the public interface clean.

**14. PR body is empty** — The commit message has good detail but the PR body is blank. For reviewers and changelog generation, the PR description should summarize what changed.

---

### Summary

| # | Severity | Area | Issue |
|---|----------|------|-------|
| 1 | **Medium** | `index.ts` | Shared `edges` singleton stored in per-reload `GraphState` — misleading ownership |
| 2 | **Medium** | `index.ts` | SidePanel not closed/refreshed on graph reload — stale node refs |
| 3 | **Medium** | `index.ts` | Poll bypasses `loadGraph` validation — may pass malformed data to `updateGraph` |
| 4 | Low | `watcher` | Only detects new files, silently ignores content changes to existing files |
| 5 | Low | `Simulation.ts` | `forceCluster` rebuilds adjacency map on every tick — unnecessary O(N+E)/tick |
| 6 | **Medium** | `vite.config.ts` | Plugin loaded in prod build — `readdirSync` may throw if dirs missing |
| 7 | Low | `Simulation.ts` | O(N²) collision — fine for now, won't scale |
| 8 | Low | `index.html` | Top-level await with no error handling |
| 9 | Low | `vite.config.ts` | Hardcoded sessions path — should support env var override |
| 10 | **Merge risk** | `.gitignore` | Deleting example data conflicts with PR #4 and removes dev bootstrap path |

The graph lifecycle and position-preserving reload are well-designed. The simulation physics improvements are solid. Main concerns are the polling validation gap (item 3), the stale SidePanel on reload (item 2), and the merge conflict risk with PR #4 (item 10). Items 1 and 6 should also be addressed before merge.

theia-panel/src/index.ts
```diff
@@ -152,55 +166,213 @@ export async function mount(
...
+      }
+    } catch (err) {
+      console.warn("[theia] poll error:", err);
+      emit("poll-error", err);
+    }
+  }
+
+  function startPolling(intervalMs = 5000) {
+    stopPolling();
+    pollTimer = setInterval(checkForUpdate, intervalMs);
```
zebster-cmd, line 328: **Polling race condition**: `setInterval` fires regardless of whether the previous `checkForUpdate` has resolved. If a fetch takes longer than `intervalMs`, two `updateGraph` calls can interleave.

Add a guard similar to the `rebuilding` flag in the watcher plugin:

```ts
let isChecking = false;

async function checkForUpdate() {
  if (disposed || isChecking) return;
  isChecking = true;
  try {
    // ... existing logic ...
  } finally {
    isChecking = false;
  }
}
```

theia-panel/vite-plugin-theia-watcher.ts
```diff
@@ -0,0 +1,98 @@
...
+      // Existing files being written to (active .jsonl) are ignored.
+      let knownFiles = new Set<string>();
+      try {
+        knownFiles = new Set(
+          readdirSync(sessionsDir).filter(
+            (f) => f.endsWith(".json") || f.endsWith(".jsonl"),
+          ),
+        );
+      } catch {
+        /* ignore */
```
zebster-cmd, line 34: **Silent failure path**: If `sessionsDir` doesn't exist, `readdirSync` throws and lands here (empty catch). But then `watch(sessionsDir, ...)` on line 75 will also throw — this time unhandled — and crash the dev server.

Either:
- Wrap the `watch()` call in a try/catch and log a warning, or
- Check `existsSync(sessionsDir)` up front and skip the watcher with a clear message.

theia-panel/vite-plugin-theia-watcher.ts
```diff
@@ -0,0 +1,98 @@
...
+            "-m",
+            "theia_core",
+            sessionsDir,
+            "-o",
+            outputPath,
+          ]);
+          console.log(
+            `\x1b[36m[theia]\x1b[0m Graph rebuilt in ${Date.now() - start}ms → ${outputPath}`,
+          );
+        } catch (e: any) {
```
zebster-cmd, line 56: Nit: `catch (e: any)` — prefer `catch (e: unknown)` and narrow with a type guard:

```ts
catch (e: unknown) {
  const msg = e instanceof Error ? e.message : String(e);
  const stderr = (e as any)?.stderr;
  console.error(`[theia] Rebuild failed:`, stderr || msg);
}
```

theia-panel/src/index.ts
```diff
@@ -152,55 +166,213 @@ export async function mount(
...
+
+    const picker = createPicker(element, ctx.camera, nodes);
+    picker.onHover(onPickerHover);
+    picker.onHover(onPickerHoverEmit);
+
+    state = {
+      graph,
+      nodes,
+      edges,
+      simulation,
```
zebster-cmd, line 232: Minor: `edges` is created once (line 65) and reused across all graph lifecycles, but it's stored in `GraphState` as `edges` — making it look like each state owns its own edge layer. Consider removing `edges` from `GraphState` since it's effectively a singleton, or add a comment clarifying the sharing.

theia-panel/src/index.ts
```diff
@@ -152,55 +166,213 @@ export async function mount(
...
+
+  // -------- polling --------
+
+  let pollTimer: ReturnType<typeof setInterval> | null = null;
+  let lastGeneratedAt: string | null = null;
+
+  async function checkForUpdate() {
+    if (disposed) return;
+    try {
+      const res = await fetch(graphUrl + "?t=" + Date.now());
```
zebster-cmd, line 308: The cache-bust `?t=Date.now()` ensures a fresh fetch, but this downloads the full graph JSON every 5 seconds. For large graphs, consider a lighter pre-check:

```ts
const headRes = await fetch(graphUrl, { method: 'HEAD' });
const lastMod = headRes.headers.get('last-modified');
```

Or use `If-Modified-Since` to get a 304 when nothing changed.

theia-panel/vite.config.ts
```diff
@@ -1,10 +1,21 @@
 import { defineConfig } from "vite";
+import { theiaWatcherPlugin } from "./vite-plugin-theia-watcher";
+import { resolve } from "path";
+import { homedir } from "os";
+
+const PROJECT_ROOT = resolve(__dirname, "..");
+const SESSIONS_DIR = resolve(homedir(), ".hermes", "sessions");
+const OUTPUT_PATH = resolve(PROJECT_ROOT, "examples", "graph.json");
+const PYTHON_PATH = resolve(PROJECT_ROOT, ".venv", "bin", "python");
```
zebster-cmd, line 9: This assumes a Unix-style `.venv/bin/python`. On Windows it would be `.venv\Scripts\python.exe`. Since this is a dev-only config, it's probably fine, but consider falling back to just `"python"` (system PATH) if the venv binary doesn't exist.

zebster-cmd (commented):
## Code Review: feat: hot-reload polling, graph lifecycle, watcher plugin

### Build Status: PASSES

The branch compiles and builds cleanly.

---

### Architecture

The refactor to a `GraphState` lifecycle (`initGraph` / `updateGraph`) with polling is well-structured. Separating the persistent scene/camera/post-processing from the per-graph nodes/simulation/picker is the right layering. The Vite watcher plugin for auto-rebuilding on new sessions is a nice DX touch.

---

### Issues

See inline comments for details. Summary:

| # | Severity | File | Issue |
|---|----------|------|-------|
| 1 | Medium | `index.ts:305-324` | Polling has no concurrency guard — overlapping fetches can race |
| 2 | Medium | `vite-plugin-theia-watcher.ts:33` | Empty `catch` swallows `readdirSync` error but `watch()` on line 75 will throw if dir is missing |
| 3 | Low | `vite-plugin-theia-watcher.ts:56` | `catch (e: any)` — prefer `unknown` |
| 4 | Low | `index.ts:65,232` | `edges` is a singleton shared across states but stored in `GraphState` as if per-graph |
| 5 | Low | `vite.config.ts:9` | `pythonPath` hardcodes Unix `.venv/bin/python`; no Windows support or fallback |
| 6 | Low | `index.ts:308` | Full graph re-fetched every poll; consider `If-Modified-Since` or `HEAD` check first |
| 7 | Info | `.gitignore` / examples | Example data deleted — fresh clone has no graph.json for `npm run dev` |
| 8 | Info | Merge conflict | PR #5 sets `port: 5179` in vite.config, this PR sets `port: 5178` — will conflict |

---

### Summary

The hot-reload architecture is solid and the graph teardown/rebuild path is clean. The main actionable items are the polling race condition and the watcher's silent failure when the sessions directory doesn't exist.

zebster-cmd:
## Merge Conflict Resolution Guide

This branch has 3 conflicts against `main` after PR #5 (`feat/3d-nodes`) was merged. Here is a file-by-file walkthrough.

### Setup

```bash
git checkout feat/live-reloading
git fetch origin main
git merge origin/main
# or: git rebase origin/main
```

---

### 1. `examples/graph.json` — modify/delete conflict

**What happened:** PR #5 updated this file, PR #6 deleted it.

**Resolution:** Since PR #6 intentionally deletes example data and adds it to `.gitignore` (generated output should not be committed), **accept the deletion**:

```bash
git rm examples/graph.json
```

---

### 2. `theia-panel/vite.config.ts` — content conflict

**What happened:** PR #5 set `port: 5179`, PR #6 set `port: 5178` and added the watcher plugin + python config.

**Resolution:** Keep PR #6's watcher plugin additions and pick one port. Since PR #5 (now on `main`) established `5179`, it is probably simplest to keep that:

- Accept all of PR #6's structural changes (watcher plugin import, `pythonPath`, `sessionsDir`, plugin array changes)
- Use `port: 5179` (what `main` now has)

---

### 3. `theia-panel/src/index.ts` — content conflict (largest)

**What happened:** PR #5 introduced the 3D star cluster transition, `SceneManager`, and new rendering pipeline. PR #6 introduced the `GraphState` lifecycle, `initGraph`/`updateGraph`, and polling hot-reload. Both heavily restructured this file.

**Resolution approach:**

- **Keep PR #5's rendering/scene changes** (these are already on `main`): `SceneManager`, Three.js scene setup, 3D node rendering, post-processing pipeline.
- **Layer PR #6's lifecycle on top**: `GraphState` type, `initGraph()` / `updateGraph()` / `disposeGraph()` functions, polling via `startPolling()` / `checkForUpdate()`.
- Key integration points to watch:
  - Make sure `initGraph` creates nodes/simulation/picker using PR #5's 3D rendering (not the old 2D path).
  - `updateGraph` should tear down and rebuild using the same 3D node creation logic from PR #5.
  - The `mount()` export should wire up PR #5's scene/camera/post-processing as the persistent layer, with PR #6's `GraphState` as the per-reload layer.

This is the most involved conflict. If it gets messy, it may be easier to:
1. Start from `main` (which has PR #5)
2. Re-apply PR #6's changes manually on top, adapting the lifecycle to the new 3D architecture

---

### After resolving

```bash
git add .
git commit   # or git rebase --continue
git push --force-with-lease
```

CI should re-run. Note: the CI failure (watcher plugin crashing in test/build) is a separate issue from the conflicts — see review comment about gating the plugin to dev-only mode.

PR #8, opened 2026-04-23
feat(reloading-v2): watch mode, DB migration, title/preview display, …

arm64be:
…edge dedup

Core:
- Add polling file watcher (theia_core/watch.py) with debounce
- Replace JSONL scanning with SQLite DB queries against Hermes state.db
- CLI now defaults to $THEIA_HOME/state.db and $THEIA_HOME/theia-graph.json
- Graph nodes include title (from DB) and preview (first user message)
- Add --watch and --watch-interval flags to auto-regenerate on DB changes
- Add reload() to panel Controller preserving camera/selection/filter state

Panel:
- Display session title in tooltip, search bar, and side panel
- Search matches against title, preview, summary, and initial_prompt
- Deduplicate edges: only highest-weight edge per kind between node pairs
- Add getCameraState/setCameraState to SceneContext

Tests:
- Rewrite tests to use temp SQLite DB seeded from fixtures
- Remove test_ingest_jsonl.py (JSONL no longer main code path)
- All 28 core tests + 2 panel tests pass, mypy and typecheck clean

Docs:
- Update README.md, build-pipeline.md, integration.md for new DB-based flow


zebster-cmd (commented):
## Code Review — PR #8: watch mode, DB migration, title/preview display

### Overall

Solid PR — the migration from JSON/JSONL file scanning to SQLite is the right call, the watch mode is clean, and the reload() controller method preserves camera/selection state nicely. The panel-side picker improvements (distance-aware radius, sticky hover, scroll/drag blocking) are good QoL. A few issues need attention before merge though.

---

### Severe Issues (inline comments posted)

1. **N+1 query in `load_sessions()`** — fires a separate `SELECT ... FROM messages` for every session row. With 500 sessions this is 501 queries. Should be a single batch query with a `GROUP BY session_id` or at minimum a single `SELECT ... ORDER BY session_id, timestamp` then partition in Python.

2. **SQLite connection not using WAL mode** — the watcher polls `state.db` mtime and calls `generate()`, which opens a new connection and reads while Hermes may be writing. Without `PRAGMA journal_mode=WAL`, this risks `SQLITE_BUSY` under concurrent access.

3. **`watch_sessions()` is dead code** — only `watch_db()` is called from `__main__.py`. The `watch_sessions()` function and `_collect_session_paths()` helper are unreachable. Either wire them up or remove them.

4. **Picker regression: `PickerOptions` interface added to diff but not used** — commit 2 adds `PickerOptions` with `shouldBlock`, `maxDistance`, `sizeScale` but the actual `createPicker()` call in `index.ts:105` on the branch doesn't pass `options`. The signature still only takes 4 args. The QoL picker commit appears to have been partially reverted or rebased incorrectly.

5. **`_extract_initial_prompt()` short-circuits on `sess.preview`** — if `sess.preview` is set (which it always is from DB sessions), it returns the 60-char truncated preview instead of the full initial prompt. This means `initial_prompt` in the graph JSON is always the truncated preview, never the full message.

6. **Contract CI workflow no longer generates `graph.json`** — the step was removed but `examples/graph.json` is still validated by `check-jsonschema`. If that file goes stale or is deleted, CI will fail. Either re-add generation from DB fixtures or commit a known-good `examples/graph.json`.

### Non-Severe Observations

- **`typing.Callable` import in `watch.py`** — uses `from typing import Callable` but also has `from __future__ import annotations`. Should use `from collections.abc import Callable` to match the rest of the codebase (this is actually inconsistent between the import and the type annotation in the same file).
- **`_build_preview()` truncation** — 60 chars is quite short. Consider making it configurable or at least a module constant.
- **`title` fallback to `session_id`** — `title = row["title"] or session_id` means an empty string title ("") also falls back. This is probably fine but worth noting.
- **Schema: `preview` allows null** — `"type": ["string", "null"]` but `Session.preview` defaults to `""` (empty string). The emit code always sets `"preview": sess.preview` which will be `""` not `null`. The schema allows both which is fine, but the code path never produces `null`.
- **`edgesScene` created but never added to `ctx.scene`** — this is intentional (it's rendered separately via `post.renderEdges()`), but worth a comment.
- **Docs table formatting** — `build-pipeline.md` line 84 has an extra column separator in the table header.

theia-core/theia_core/ingest.py
```diff
@@ -161,6 +162,106 @@ def _infer_search_hits(
...
+    for msg in messages:
+        if msg.get("role") == "user":
+            content = msg.get("content", "")
+            if isinstance(content, str) and content.strip():
+                stripped = content.strip()
+                text = stripped[:60]
+                return text + ("..." if len(stripped) > 60 else "")
+    return ""
+
+
```
zebster-cmd, line 179: **SEVERE: N+1 query pattern.** This fires a separate `SELECT ... FROM messages WHERE session_id = ?` for every session row. With 500 sessions that's 501 round-trips to SQLite.

Fix: fetch all messages in one query ordered by `session_id, timestamp`, then partition in Python:
```python
all_msgs = conn.execute(
    "SELECT session_id, role, content, tool_calls, tool_call_id, timestamp "
    "FROM messages WHERE session_id IN (SELECT id FROM sessions WHERE parent_session_id IS NULL) "
    "ORDER BY session_id, timestamp, id"
).fetchall()
msg_by_session: dict[str, list[dict]] = {}
for row in all_msgs:
    msg_by_session.setdefault(row["session_id"], []).append(dict(row))
```
Then look up `msg_by_session.get(session_id, [])` in the session loop.

theia-core/theia_core/ingest.py
```diff
@@ -132,7 +134,6 @@ def _infer_search_hits(
         except json.JSONDecodeError:
             args = {}
         query = str(args.get("query", ""))
-        # Try arguments first, then tool response results
```
zebster-cmd, line 135: **SEVERE: Missing WAL mode.** This connection reads `state.db` while Hermes may be actively writing to it. Without WAL, concurrent readers and writers will hit `SQLITE_BUSY`.

Add after `sqlite3.connect()`:
```python
conn.execute("PRAGMA journal_mode=WAL")
```
Also consider a short `busy_timeout`:
```python
conn.execute("PRAGMA busy_timeout=3000")
```

theia-core/theia_core/watch.py
```diff
@@ -0,0 +1,101 @@
...
+
+def _snapshot(paths: list[Path]) -> dict[Path, float | None]:
+    """Return mtime for each path; None if the path does not exist."""
+    snap: dict[Path, float | None] = {}
+    for p in paths:
+        try:
+            snap[p] = p.stat().st_mtime
+        except OSError:
+            snap[p] = None
+    return snap
```
zebster-cmd, line 16: **SEVERE: Dead code.** `watch_sessions()` and `_collect_session_paths()` are never called — `__main__.py` only imports and uses `watch_db()`. This is ~40 lines of unreachable code.

Either:
1. Remove both functions (the file-based session watching is replaced by DB watching)
2. Or expose `watch_sessions` in `__init__` and document when it should be used

theia-core/theia_core/emit.py
```diff
@@ -24,13 +24,14 @@ def _extract_summary(sess: Session) -> str | None:
 
 
 def _extract_initial_prompt(sess: Session) -> str | None:
+    """Return the first user message content, or None."""
+    if sess.preview:
```
zebster-cmd: **SEVERE: `_extract_initial_prompt()` always returns the truncated preview instead of the full prompt.** 

The `sess.preview` field is always populated for DB sessions (it's the first 60 chars of the first user message). This early return means `initial_prompt` in the graph JSON is always the 60-char truncated string, never the full message content.

This defeats the purpose of having both `preview` (short) and `initial_prompt` (full). Either:
1. Remove the `sess.preview` short-circuit and let the existing logic find the full message
2. Or keep it, but don't truncate `preview` at 60 chars

theia-panel/src/index.ts
```diff
@@ -32,32 +33,65 @@ export async function mount(
...
+  let simNodes: ReturnType<typeof createSimulation>["nodes"] = [];
+  let simulation: ReturnType<typeof createSimulation>["simulation"];
+  let picker: ReturnType<typeof createPicker>;
+  let searchBar: ReturnType<typeof createSearchBar>;
+
+  const tooltip = createTooltip(element);
+  const sidePanel = createSidePanel(element);
 
-  const { simulation, nodes: simNodes } = createSimulation(graph);
-  simulation.stop();
```
zebster-cmd, line 56: **SEVERE: Picker options not wired up.** Commit 2 adds `PickerOptions` (with `shouldBlock`, `maxDistance`, `sizeScale`) and the `pickAt()` method, but `setupGraph()` in index.ts calls:
```ts
picker = createPicker(element, ctx.camera, nodes, nodePositions);
```
without passing options. The `shouldBlock` callback and distance-aware radius are never actually used. The `isInteracting()` helper and `lastWheelAt` tracking are set up but disconnected.

The `pickAt()` call on mouseup also references `pickNDC` with `radiusScale` but the base picker doesn't have `pickAt` in the 4-arg overload.

Looks like commit 2 was partially reverted during the rebase. Need to pass the options through:
```ts
picker = createPicker(element, ctx.camera, nodes, nodePositions, {
  shouldBlock: isInteracting,
});
```

.github/workflows/contract.yml
```diff
@@ -18,10 +18,7 @@ jobs:
       - name: Install core
         working-directory: theia-core
         run: pip install -e ".[dev]"
-      - name: Generate graph.json from fixtures
-        run: |
```
zebster-cmd, line 22: **SEVERE: CI will break if `examples/graph.json` goes stale.** The step that generated `examples/graph.json` from fixtures was removed, but the next step still validates it:
```yaml
check-jsonschema --schemafile schemas/graph.schema.json examples/graph.json
```
This relies on a previously-committed `examples/graph.json` that was generated from the old file-based CLI. If anyone updates the schema (e.g. `preview` is now required, or `title` becomes optional), the stale file will fail validation.

Either:
1. Re-add generation using the DB-based CLI (seed a temp DB from fixtures, run `python -m theia_core --db-path ... -o examples/graph.json`)
2. Or remove the validation step and rely on `test_cli_produces_valid_graph` instead

PR #10, opened 2026-04-24
Feat/subagent sessions

arm64be:
Summary
- Fixes subagent node visibility tracking (scale vs visibility conflation)
- Fixes subagent nodes hidden when connected via non-subagent edges
- Guards side-panel edge lists against active filter state
Changes
Nodes.ts — setVisible was encoding visibility into the instance matrix scale (0 = hidden). setPosition decomposed the matrix and blindly preserved whatever scale it found, so hidden nodes stayed hidden through position updates. Added a visibleFlags array; setPosition now always consults it to compute the correct scale, decoupling the two concerns.
index.ts (computeVisibleNodeIds) — When subagent filter is off, subagent nodes that participate in non-subagent edges (e.g. memory-share) were still excluded because the edge loop only marked non-subagent endpoints. Now those nodes are correctly included.
index.ts (side panel / search) — Edge lists shown in the side panel and search results now respect the active kinds filter.
Remaining issue (needs eyes)
Subagent-dot balloons still don't render when the "subagent" filter is on. The subagent edges draw fine (soft green lines), and the subagent nodes are hoverable/pickable — so they exist in the scene and the picker can find them via nodePositions + isVisible. But the glowing sprite ball itself is invisible.
I've ruled out:
- Scale/visibility conflation (fixed above — visibleFlags is consulted in setPosition)
- computeVisibleNodeIds returning wrong IDs (returns all nodes when subagent enabled)
- nodePositions being stale (updated every frame in tick())
- Post-processing compositing (edges render fine in same pipeline)
- Shader distance fade (smoothstep(0.15, 0.8, dist) — camera is ~10 units away, so fade ≈ 1)
- Depth testing (depthWrite: false, transparent: true, NormalBlending)
The picker does its own NDC math against nodePositions — it doesn't use Three.js raycasting — which explains why it finds nodes the mesh doesn't render. Something in the instance matrix or render path is silencing the draw. Likely candidates: stale instanceMatrix.needsUpdate, simulation writing NaN positions, or a blending/depth interaction I'm not seeing.
Would appreciate a second set of eyes on tick() → setPosition() → flush() flow and the fragment shader.


zebster-cmd (changes requested):
## Code Review: PR #10 — `feat/subagent-sessions`

**Branch:** `feat/subagent-sessions` → `main` | **+938 / -28** across 18 files

---

### Overview

This PR adds subagent (parent–child) session detection to both the `theia-core` Python pipeline and the `theia-panel` Three.js visualization. It introduces a new `"subagent"` edge kind, a `parent_id` field on graph nodes, filter-toggle visibility for child sessions, and side-panel detail rendering.

The change is well-structured with clean separation between the core detector and panel visualization layers. Two commits: the initial feature implementation and a follow-up fix for a visibility/scale conflation bug.

---

### Issues

#### 1. Bug: `computeVisibleNodeIds` does not match stated behavior

**File:** `theia-panel/src/index.ts` — `computeVisibleNodeIds()`

The commit message (`b9360a49`) states: *"when subagent filter is disabled, subagent nodes that have non-subagent connections (e.g. memory-share) are now treated as visible"*.

However, the code does **not** achieve this. Subagent children are **never** added to `hasNonSubagentConnection` because the edge loop explicitly skips them:

```typescript
if (!subagentIds.has(edge.source)) {
  hasNonSubagentConnection.set(edge.source, true);  // subagent children skipped
}
if (!subagentIds.has(edge.target)) {
  hasNonSubagentConnection.set(edge.target, true);  // subagent children skipped
}
```

A subagent child (`parent_id` set) that also has `memory-share` or `cross-search` edges will still be hidden when the subagent filter is off. To match the documented intent, remove the `!subagentIds.has()` guard:

```typescript
// Both endpoints should be marked, regardless of subagent status
hasNonSubagentConnection.set(edge.source, true);
hasNonSubagentConnection.set(edge.target, true);
```

#### 2. Type unsafety: `ev.child_session_id` access on opaque evidence type

**File:** `theia-panel/src/ui/SidePanel.ts`

```typescript
const displayId = isChild ? e.source : ev.child_session_id;
```

The `evidence` field is typed as `{}` in `theia-panel/src/data/types.ts`, so `ev.child_session_id` is a type error under `noPropertyAccessFromIndexSignature` or strict mode. Either:
- Widen the evidence type to `Record<string, unknown>`, or
- Cast explicitly: `(ev as Record<string, unknown>).child_session_id`

#### 3. `ingest.py` — behavioral breaking change for `load_sessions()`

**File:** `theia-core/theia_core/ingest.py`

Removing `WHERE parent_session_id IS NULL` means `load_sessions()` now returns **all** sessions (including subagent children), not just top-level ones. This is correct for Theia's purposes, but if any downstream consumers relied on the old behavior (only top-level), they'd silently get extra sessions. Consider adding an optional `include_children: bool = True` parameter for backward compatibility, or document the change prominently.

---

### Nits / Minor

#### 4. Double `simulation.stop()` in `updateVisibility`

**File:** `theia-panel/src/index.ts` — `updateVisibility()`

`createSimulation` returns a running simulation, then `simulation.stop()` is called, positions are copied, and `simulation.stop()` is called again. The second stop is redundant.

#### 5. `detect_subagent.py` — consider validating parent existence

**File:** `theia-core/theia_core/detect/subagent.py`

The detector creates edges from `sess.parent_id → sess.id` without checking whether `parent_id` corresponds to an existing session in the list. The edge filter in `emit.py` catches dangling edges silently, but an explicit guard would be clearer:

```python
session_ids = {s.id for s in sessions}
for sess in sessions:
    if sess.parent_id and sess.parent_id in session_ids:
        ...
```

#### 6. Regeneration cost on every filter toggle

Every filter toggle reconstructs the full force simulation, copies all positions, and rebuilds all edges. For small-to-medium graphs this is fine, but worth noting as a scaling concern. A lighter approach would be to just update edge/node visibility without reinitializing the physics.

#### 7. Lock file noise

**File:** `theia-core/uv.lock` — +754 lines

The auto-generated lock file dominates the diff. Consider marking it via `.gitattributes`:
```
uv.lock linguist-generated=true
```

---

### What's Good

- Clean new detector module (`detect/subagent.py`) — 21 lines, single responsibility, well-typed
- Schema and type definitions updated consistently across Python and TypeScript
- The `visibleFlags` array in `Nodes.ts` properly decouples visibility state from instance matrix scale — the right fix for the Three.js InstancedMesh conflation issue
- Search, picker, and side panel all respect the new visibility filter
- Soft green palette choice (`#7ce38b`) contrasts well against the existing amber/cyan/violet

---

### Verdict

**Request changes** on item #1 (the `computeVisibleNodeIds` bug contradicts the documented fix). Items #2–#3 are worth addressing before merge; the rest are optional nits.

zebster-cmd:
## Follow-up: Subagent node sprites invisible when filter is ON

The subagent edges render correctly (soft green lines) and the nodes are hoverable/pickable via `nodePositions` + `isVisible`, but the actual glowing sprite ball for subagent nodes never appears visually.

This points to a rendering issue in `Nodes.ts` rather than a data or visibility-logic issue. Here is where I'd investigate:

### Likely culprit: `setVisible` / `setPosition` scale interaction

When `setVisible(i, true)` is called, it sets the instance scale to `nodeSizes[i]`:

```typescript
setVisible(i, visible) {
  visibleFlags[i] = visible;
  mesh.getMatrixAt(i, dummy.matrix);
  dummy.matrix.decompose(dummy.position, dummy.quaternion, dummy.scale);
  const sz = visible ? nodeSizes[i]! : 0;
  dummy.scale.set(sz, sz, sz);
  dummy.updateMatrix();
  mesh.setMatrixAt(i, dummy.matrix);
}
```

But `setPosition` **also** applies scale based on `visibleFlags`:

```typescript
setPosition(i, x, y, z) {
  mesh.getMatrixAt(i, dummy.matrix);
  dummy.matrix.decompose(dummy.position, dummy.quaternion, dummy.scale);
  dummy.position.set(x, y, z);
  const sz = visibleFlags[i] ? nodeSizes[i]! : 0;
  dummy.scale.set(sz, sz, sz);
  dummy.updateMatrix();
  mesh.setMatrixAt(i, dummy.matrix);
}
```

The problem is **call ordering during `setupGraph`**. Here's the sequence:

1. `createNodes(g)` initializes all instances with their `nodeSizes` scale ✅
2. The simulation tick loop calls `setPosition()` for each node, which decomposes the matrix and re-applies scale from `visibleFlags` ✅
3. `computeVisibleNodeIds` is called — when subagent filter is ON, all nodes should be visible ✅
4. `setVisible(i, true)` is called for each node ✅
5. `nodes.flush()` is called ✅

But consider: **`visibleFlags` is initialized as `new Array(n).fill(true)`** inside `createNodes`. When `setupGraph` calls `computeVisibleNodeIds` and then `setVisible`, it's operating on the *current* `nodes` object. However, `setupGraph` **recreates** `nodes` via `createNodes(g)` at the top — so the `visibleFlags` array inside the new `nodes` object starts fresh as all-`true`.

The real question is: **what happens on the animation frame tick?** The render loop calls `setPosition` for every node on every simulation tick. If `setPosition` is called **before** `setVisible` has been called after a filter change (via `updateVisibility`), the stale `visibleFlags` (all `false` from a previous state where subagent was off) would set scale to 0.

**Check whether `updateVisibility()` is racing with the animation loop.** If the render tick fires between `computeVisibleNodeIds` and the `setVisible` calls, nodes could be drawn at scale 0 and then `flush()` never gets called again to fix them.

### Additional things to check

1. **`nodeSizes[i]` for subagent nodes** — Are subagent child sessions producing a valid `nodeSizes` value? If `nodeSizes[i]` is `0` or `undefined` for a subagent node (e.g., because it has zero tool calls or zero messages), then even `visible ? nodeSizes[i]! : 0` would result in scale 0.

2. **`instanceMatrix.needsUpdate`** — After `setVisible` + `flush`, is `mesh.instanceMatrix.needsUpdate = true` being called? Check that `flush()` actually triggers a GPU upload.

3. **Instance color for subagent nodes** — Even if the scale is correct, if `setTime()` or the color initialization sets the color to fully transparent or black for subagent nodes, they'd be invisible. Check the color computation in `createNodes` — does it handle `parent_id`-bearing nodes differently?

The most likely fix is either a race condition in the animation loop or `nodeSizes[i]` being zero for low-activity subagent sessions.

zebster-cmd:
## Root Cause Found: NaN poisoning via `Matrix4.decompose()` on zero-scale matrices

Investigated locally. The bug is in `theia-panel/src/scene/Nodes.ts` — specifically the `setVisible()` and `setPosition()` methods both use a `getMatrixAt → decompose → modify → updateMatrix → setMatrixAt` roundtrip that breaks when the intermediate matrix has scale=0.

### The smoking gun

```typescript
setVisible(i, visible) {
  visibleFlags[i] = visible;
  mesh.getMatrixAt(i, dummy.matrix);
  dummy.matrix.decompose(dummy.position, dummy.quaternion, dummy.scale);  // <-- BUG
  const sz = visible ? nodeSizes[i]! : 0;
  dummy.scale.set(sz, sz, sz);
  dummy.updateMatrix();
  mesh.setMatrixAt(i, dummy.matrix);
}
```

### Why this breaks

When `setVisible(i, false)` runs for a subagent child during initial `setupGraph` (subagent filter is OFF by default), the instance matrix gets scale = 0 in all three columns.

When the user later toggles subagent ON, `updateVisibility` calls `setVisible(i, true)`:

1. `mesh.getMatrixAt(i, dummy.matrix)` — matrix has scale 0
2. `dummy.matrix.decompose(...)` — **this is where it breaks**

Three.js `Matrix4.decompose()` computes `invSX = 1 / sx` where `sx = length(col0) = 0`, giving `invSX = Infinity`. It then multiplies each rotation-part element by `Infinity`: `0 * Infinity = NaN`. This produces a **NaN rotation matrix**, and `quaternion.setFromRotationMatrix()` turns that into a **NaN quaternion**.

3. `dummy.scale.set(nodeSizes[i], ...)` — scale is valid
4. `dummy.updateMatrix()` calls `compose(position, quaternion, scale)` — with a **NaN quaternion**, every rotation+scale element comes out NaN. Translation (`te[12..14]`) is still valid.
5. `setMatrixAt(i, dummy.matrix)` writes the poisoned matrix.

### Why it matches the symptoms exactly

Look at the node vertex shader:

```glsl
vec3 instancePos = instanceMatrix[3].xyz;            // translation — still VALID
vec3 scale = vec3(
  length(instanceMatrix[0].xyz),                     // length(NaN vec3) = NaN
  length(instanceMatrix[1].xyz),
  length(instanceMatrix[2].xyz)
);
vec4 viewPos = viewMatrix * vec4(instancePos, 1.0);  // still valid
viewPos.xy += position.xy * scale.xy;                // NaN
gl_Position = projectionMatrix * viewPos;            // NaN → GPU clips
```

This explains **every** observation:

| Observation | Explanation |
|---|---|
| Edges render fine | `Edges.ts` writes to `nodePositions` buffer; never touches `instanceMatrix` |
| Picker finds nodes | `Picker.ts` reads `nodePositions`; also never touches `instanceMatrix` |
| Sprite is invisible | Vertex shader `gl_Position = NaN`, GPU clips all verts |
| Translation survives but scale-part is NaN | `compose()` uses valid `position` but NaN `quaternion` |

The decompose/compose roundtrip is entirely unnecessary — the node quads never have meaningful rotation (the shader only reads translation and column lengths). The "preserve existing rotation" intermediate step is what poisons the state.

### Fix options

**Option 1 (minimal):** Reset quaternion to identity before `updateMatrix()`:

```typescript
setVisible(i, visible) {
  visibleFlags[i] = visible;
  mesh.getMatrixAt(i, dummy.matrix);
  dummy.matrix.decompose(dummy.position, dummy.quaternion, dummy.scale);
  dummy.quaternion.identity();   // <-- avoid NaN propagation
  const sz = visible ? nodeSizes[i]! : 0;
  dummy.scale.set(sz, sz, sz);
  dummy.updateMatrix();
  mesh.setMatrixAt(i, dummy.matrix);
}
```
(Same for `setPosition`.)

**Option 2 (cleaner):** Skip decompose entirely — track positions in a `Float32Array` and always rebuild the matrix from known state:

```typescript
const positions = new Float32Array(n * 3);

setPosition(i, x, y, z) {
  positions[i*3] = x;
  positions[i*3+1] = y;
  positions[i*3+2] = z;
  const sz = visibleFlags[i] ? nodeSizes[i]! : 0;
  dummy.position.set(x, y, z);
  dummy.quaternion.set(0, 0, 0, 1);
  dummy.scale.set(sz, sz, sz);
  dummy.updateMatrix();
  mesh.setMatrixAt(i, dummy.matrix);
},

setVisible(i, visible) {
  visibleFlags[i] = visible;
  const sz = visible ? nodeSizes[i]! : 0;
  dummy.position.set(positions[i*3], positions[i*3+1], positions[i*3+2]);
  dummy.quaternion.set(0, 0, 0, 1);
  dummy.scale.set(sz, sz, sz);
  dummy.updateMatrix();
  mesh.setMatrixAt(i, dummy.matrix);
}
```

**Option 3 (best):** `nodePositions` already exists at the `index.ts` level — reuse it in the layer rather than duplicating.

PR #12, opened 2026-04-24
style: adapt panel and plugin to match Hermes Agent dashboard theming

arm64be:
- SidePanel: replace hardcoded translucent backgrounds with theme-aware cardBg helper (blends fg into bg at 4%, matching dashboard bg-card pattern)
- SidePanel: redesign edge cards with compact badge-style kind labels, uppercase letter-spaced dt labels, subtle section dividers
- SidePanel: adjust typography (line-height, font-family, letter-spacing) to align with dashboard conventions
- plugin/style.css: replace hardcoded #07080d with var(--background-base) and add var(--radius), var(--midground), var(--theme-font-sans) so the plugin shell reflects the active dashboard theme

### DIFF
--- a/plugin/src/index.js
+++ b/plugin/src/index.js
@@ -77,24 +77,26 @@
     function cssVar(name) {
       return (root.getPropertyValue(name) || "").trim();
     }
-    var bg     = toHex(cssVar("--background-base"), "07080d");
+    var bg       = toHex(cssVar("--background-base"), "07080d");
     // fg — primary body text: use the midground triplet (theme-responsive).
-    var fg     = toHex(cssVar("--midground-base"), "cfd6e4");
+    var fg       = toHex(cssVar("--midground-base"), "cfd6e4");
     // fg2 — secondary/muted labels (55% opacity midground in the DS).
-    var fg2    = toHex(cssVar("--color-muted-foreground") || cssVar("--midground"), "9ca3af");
+    var fg2      = toHex(cssVar("--color-muted-foreground") || cssVar("--midground"), "9ca3af");
+    // midground — the dashboard's midground layer used e.g. for bg-card blend.
+    var midground = toHex(cssVar("--midground-base"), "cfd6e4");
     // accent — warm gold for emphasis.  --color-warning is static (#ffbd38)
     // but intentionally fixed: it provides a functional highlight that stands
     // out from body text across all dark themes.  Falls back to the theme's
     // midground only if --color-warning is absent.
-    var accent = toHex(cssVar("--color-warning") || cssVar("--midground-base"), "ffc477");
-    var border = toHex(cssVar("--color-border"), "ffffff26");
-    var font   = root.getPropertyValue("font-family").trim()
+    var accent   = toHex(cssVar("--color-warning") || cssVar("--midground-base"), "ffc477");
+    var border   = toHex(cssVar("--color-border"), "ffffff26");
+    var font     = root.getPropertyValue("font-family").trim()
       || "'Mondwest', ui-monospace, monospace";
-    return { bg: bg, fg: fg, fg2: fg2, accent: accent, border: border, font: font, radius: "0" };
+    return { bg, fg, fg2, midground, accent, border, font, radius: "0" };
   }
 
   function buildThemeQuery(theme) {
-    var keys = ["bg", "fg", "fg2", "accent", "border", "font", "radius"];
+    var keys = ["bg", "fg", "fg2", "midground", "accent", "border", "font", "radius"];
     return keys.map(function (k) {
       return k + "=" + encodeURIComponent(theme[k]);
     }).join("&");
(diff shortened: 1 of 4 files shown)


zebster-cmd:
## Code Review

### Summary

This PR replaces hardcoded colors and fonts with theme-aware equivalents across the plugin CSS and the side panel TypeScript. The core approach is sound: the new `cardBg()` function in `SidePanel.ts` does programmatic color blending to match the dashboard's `bg-card` pattern, and the CSS changes in `style.css` swap magic values for CSS custom properties with sensible fallbacks. The visual redesign (badge-style kind labels, uppercase dt labels, section dividers) is coherent and aligns with the dashboard's design language. However, there are robustness gaps in the hex parser, a questionable CSS variable name, an accessibility regression, and a subtle UX loss on edge cards.

### Issues

#### Major

**1. `cardBg()` has no guard against malformed hex — produces `rgb(NaN,NaN,NaN)`**
`theia-panel/src/ui/SidePanel.ts:9-19` — If `t.fg` or `t.bg` are ever empty, shorter than 6 characters, or prefixed with `#`, every `parseInt` returns `NaN`, and the function emits the CSS value `rgb(NaN,NaN,NaN)`, which browsers ignore silently — leaving the card with a transparent background (no visible card at all). The existing `themeBgAlpha()` at `Theme.ts:86` already does `hex.replace(/^#/, "")` defensively; `cardBg()` does not.

A one-line guard and a `NaN` fallback would make this resilient:

```ts
function cardBg(t: ThemeTokens): string {
  const parse = (hex: string) => {
    const h = hex.replace(/^#/, "");
    return [
      parseInt(h.slice(0, 2), 16) || 0,
      parseInt(h.slice(2, 4), 16) || 0,
      parseInt(h.slice(4, 6), 16) || 0,
    ];
  };
  const [fr, fgG, fb] = parse(t.fg);
  const [br, bgG, bb] = parse(t.bg);
  const mix = (a: number, b: number, p: number) => Math.round(a * p + b * (1 - p));
  return `rgb(${mix(fr, br, 0.04)},${mix(fgG, bgG, 0.04)},${mix(fb, bb, 0.04)})`;
}
```

**2. `--midground` CSS variable is semantically wrong for the toolbar text color**
`plugin/src/style.css:56` — `color: var(--midground, #cfd6e4)`. In the plugin's own `extractDashboardTheme()` (`plugin/src/index.js:84`), `--midground` is used as a fallback for `fg2` (secondary/muted text). The fallback `#cfd6e4` here is the *primary* text default (`fg`). This is contradictory: either the variable should be `--midground-base` (which maps to `fg` primary at `index.js:82`), or the fallback should be `9ca3af` (the `fg2` default). Recommend changing to:

```css
color: var(--midground-base, #cfd6e4);
```

#### Minor

**3. `--theme-font-sans` is unverified — no evidence it exists in the dashboard**
`plugin/src/style.css:57` — This variable appears nowhere in the codebase. If the dashboard doesn't define `--theme-font-sans`, the fallback `ui-monospace, monospace` always wins, making the variable declaration dead code.

**4. Edge card weight label lost its `w=` prefix — bare numbers lack context**
A user seeing "0.73" in the edge card has no indication it represents a weight. Consider keeping the prefix or using a `title` attribute.

**5. Accessibility: `dt` labels at opacity 0.45 likely fail WCAG AA contrast**
`theia-panel/src/ui/SidePanel.ts:90-94` — The `dt` labels are styled with `opacity:0.45; color:#${theme.fg2}; font-size:10px`. With defaults (`fg2=9ca3af` ≈ RGB 156,163,175 and `bg=07080d` ≈ RGB 7,8,13), the effective contrast ratio is approximately **2.8:1** — below the WCAG AA minimum of **4.5:1** for normal text. Recommend raising to at least `0.6`.

**6. Close button opacity 0.5 with `transition:opacity` but no `:hover` state**
Inline styles can't do pseudo-classes — the hover effect won't work as likely intended.

**7. `color-mix()` browser support floor undocumented**
`plugin/src/style.css:55` — Requires Chrome 111+, Firefox 113+, Safari 16.2+. Almost certainly fine for the dashboard target, but worth a comment noting the minimum.

#### Nit

**8. Variable shadowing in `cardBg()`** — `const fg = parseInt(t.fg.slice(2, 4), 16)` uses `fg` for the *green channel*, which shadows the conceptual meaning used elsewhere. Suggest `fgG`/`bgG`.

**9. Repeated `dt` style string — 5 identical copies** — Extract to a local variable.

**10. "No connections" empty state at opacity 0.35** — Very faint on dark backgrounds.

### Verdict: Request Changes

Issues 1 and 2 should be addressed before merge:
1. **`cardBg()` robustness** — add `#` stripping and `NaN` guard (trivial fix, prevents a blank-card failure mode)
2. **`--midground` vs `--midground-base`** — use the correct variable to match the existing convention

The accessibility concern (issue 5) is a strong recommendation but may be acceptable depending on the project's accessibility stance.

PR #13, opened 2026-04-24
feat(panel): add model filter, clickable session links, keyboard nav …

arm64be:
…in search, and usability improvements

- Model filter dropdown in filter bar to filter nodes by model
- Clickable session IDs in side panel connections to navigate between sessions
- Keyboard navigation (arrow keys + enter + escape) in search bar
- Auto-scroll to selected search result, result count display
- Side panel scrolls to top when switching sessions
- onNavigate callback wired through index.ts to focus graph and refresh panel

### DIFF
--- a/theia-panel/src/index.ts
+++ b/theia-panel/src/index.ts
@@ -71,7 +71,21 @@ export async function mount(
   let searchBar: ReturnType<typeof createSearchBar>;
 
   const tooltip = createTooltip(element, theme);
-  const sidePanel = createSidePanel(element, theme);
+  const sidePanel = createSidePanel(element, theme, (targetId) => {
+    const idx = nodeIndex.get(targetId);
+    if (idx !== undefined && visibleNodeIds.has(targetId)) {
+      const sn = simNodes[idx]!;
+      ctx.focusOn(sn.x, sn.y, 1.5);
+      const n = currentGraph.nodes[idx]!;
+      const related = currentGraph.edges.filter(
+        (e) =>
+          (e.source === n.id || e.target === n.id) &&
+          kinds.has(e.kind) &&
+          visibleNodeIds.has(e.source === n.id ? e.target : e.source),
+      );
+      sidePanel.show(n, related, currentGraph);
+    }
+  });
 
   let lastMouse = { x: 0, y: 0 };
   let lastWheelAt = 0;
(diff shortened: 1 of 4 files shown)


zebster-cmd:
## Code Review

### Summary

This PR adds four features to the theia-panel:

1. **Model filter dropdown** in `FilterBar` — a `<select>` element to filter visible nodes by model, intersected with the existing edge-kind visibility logic.
2. **Clickable session IDs** in `SidePanel` connection cards — `data-navigate-to` attributes with click/keydown event delegation.
3. **Keyboard navigation** in `SearchBar` — ArrowUp/Down/Enter/Escape handling, `selectedIndex` tracking, `select()`/`commit()`/`hide()` helpers.
4. **UX polish** — result count display, auto-scroll to selected, side panel scrolls to top on session switch, `kindColor` hoisted to module scope.

The architecture is sound: the `FilterState` interface cleanly bundles kind toggles + model filter, the `onNavigate` callback is properly wired through `index.ts` using event delegation, and keyboard nav follows standard patterns. However, there are state-synchronization bugs and a behavioral regression that need attention.

### Issues

#### Major

**1. [Bug] `rebuildModelSelect()` does not reset `selectedModel` when the chosen model disappears from a new graph**
`theia-panel/src/ui/FilterBar.ts` — `rebuildModelSelect()` and `updateGraph()`

When `reload()` is called in `index.ts`, `filterBar.updateGraph(newGraph)` triggers `rebuildModelSelect()` which tears down and recreates the `<select>`. However, `selectedModel` is never cleared. If the previously-selected model doesn't exist in the new graph, the dropdown visually shows "All Models" (the old `<option>` is gone), but `selectedModel` still holds the stale model name. The next `updateVisibility()` will return an empty set — hiding all nodes with no indication of why.

**Fix:**
```ts
function rebuildModelSelect() {
  // ... existing teardown ...
  const models = new Set<string>();
  for (const node of currentGraph.nodes) {
    if (node.model) models.add(node.model);
  }
  if (selectedModel && !models.has(selectedModel)) {
    selectedModel = null;
    emitChange();
  }
  if (models.size <= 1) return;
  // ... rest unchanged ...
}
```

**2. [Bug] `rebuildModelSelect()` doesn't restore the `<select>` value after rebuild**
`theia-panel/src/ui/FilterBar.ts` — even when `selectedModel` is still valid after a graph update, the new `<select>` defaults to the first option ("All Models"). The dropdown visual doesn't match internal state.

**Fix:** After appending all options:
```ts
if (selectedModel) select.value = selectedModel;
```

**3. [Regression] `commit()` in SearchBar no longer clears `input.value`**
`theia-panel/src/ui/SearchBar.ts` — the old mouse-click handler cleared the input text after selection (`input.value = ""`). The new `commit()`, which is now the single codepath for both mouse and keyboard selection, does not. Re-focusing the input then re-renders stale results.

#### Minor

**4. [Inconsistency] `onNavigate` edge filtering is stricter than direct node-click**
`theia-panel/src/index.ts` — the `onNavigate` callback filters connections to only show edges whose other endpoint is also visible. The normal mouse-click handler does not apply this extra filter. Navigating via the side panel link shows fewer connections than clicking the same node directly.

**5. [UX] Dropdown select is destroyed mid-interaction if `reload()` fires while open**
`rebuildModelSelect()` `.remove()`'s and recreates the element, abruptly closing any open dropdown.

**6. [UX] No empty-state feedback when model filter + kind filter produces zero nodes**
The graph shows nothing with no message or indicator.

**7. [A11y] `data-navigate-to` elements lack visible focus styling**
`theia-panel/src/ui/SidePanel.ts` — `<span data-navigate-to="..." tabindex="0" role="link">` has no `:focus` or `:focus-visible` outline.

#### Nit

**8.** ArrowUp at index 0 stays at 0 — can't return to "no selection" state. Acceptable, most search UIs do this.

**9.** `separator` background uses `theme.fg` at 15% opacity rather than `theme.border` — inconsistent with the rest of the component.

**10.** Color values are scattered in `renderEdge()` detail blocks (e.g., `#66d9ef`, `#b089ff`) — could reference the new module-scope `kindColor[e.kind]`.

**11.** `onNavigate?` is optional but click/keydown listeners always register — clicking a link silently does nothing if no callback was passed (cursor/underline still suggest interactivity). Consider conditionally adding `data-navigate-to` only when `onNavigate` is provided.

### Verdict: Request Changes

Issues 1, 2, and 3 are user-visible bugs that will cause broken behavior (blank graph, stale dropdown state, persistent search query). They should be addressed before merge.

PR #14, opened 2026-04-24
feat: add one-command install script for beginners

arm64be:
The install.sh script provides a beginner-friendly way to install Theia:
- Checks prerequisites (Python 3.11+, Node.js, npm, git, make)
- Clones repo to ~/.hermes/hermes-theia/ (skips if already present)
- Installs Python & Node dependencies
- Builds the panel embed and assembles the plugin
- Symlinks plugin into ~/.hermes/plugins/theia-constellation
- Optionally installs a watch service (systemd or shell script)

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -5,24 +5,57 @@
 # Usage:
 #   curl -fsSL https://raw.githubusercontent.com/arm64be/theia/main/install.sh | bash
 #   bash install.sh                          (from a local clone)
+#   bash install.sh --help                   (show this help)
+#   bash install.sh --no-service             (skip service prompt)
+#   bash install.sh --no-update              (don't git pull if already installed)
 #
 # What it does:
-#   1. Checks prerequisites (Python ≥3.11, Node.js, npm, git, make)
-#   2. Clones the repo into ~/.hermes/hermes-theia/ (only if missing)
-#   3. Installs Python & Node dependencies
-#   4. Builds the panel embed and assembles the plugin
-#   5. Symlinks the plugin into ~/.hermes/plugins/theia-constellation
-#   6. Optionally installs a watch service (systemd or shell script)
+#   1. Checks prerequisites (Python >= 3.11, Node.js, npm, git, make)
+#   2. Clones the repo into ~/.hermes/hermes-theia/ (git pull to update if exists)
+#   3. Creates a Python virtual environment and installs dependencies
+#   4. Installs Node dependencies
+#   5. Builds the panel embed and assembles the plugin
+#   6. Symlinks the plugin into ~/.hermes/plugins/theia-constellation
+#   7. Optionally installs a watch service (systemd or shell script)
 # ---------------------------------------------------------------------------
 set -euo pipefail
 
-SKIP_SERVICE=false
-for arg in "$@"; do [ "$arg" = "--no-service" ] && SKIP_SERVICE=true; done
-
+# ---------------------------------------------------------------------------
+# Config
+# ---------------------------------------------------------------------------
 REPO_URL="https://github.com/arm64be/theia"
 INSTALL_DIR="${HOME}/.hermes/hermes-theia"
+VENV_DIR="${INSTALL_DIR}/.venv"
 HERMES_PLUGINS_DIR="${HOME}/.hermes/plugins"
 PLUGIN_NAME="theia-constellation"
+
+# ---------------------------------------------------------------------------
+# Argument parsing
+# ---------------------------------------------------------------------------
+HELP=false
+NO_UPDATE=false
+SKIP_SERVICE=false
+
+for arg in "$@"; do
+    case "$arg" in
+        --help|-h) HELP=true ;;
+        --no-update) NO_UPDATE=true ;;
+        --no-service) SKIP_SERVICE=true ;;
+    esac
+done
+
+usage() {
+    cat << 'EOF'
+Usage: bash install.sh [options]
+
+Options:
+  --help, -h       Show this help message
+  --no-service     Skip the watch service prompt
+  --no-update      Don't git pull if already installed
+EOF
+}
+
+$HELP && { usage; exit 0; }
 
 # ---------------------------------------------------------------------------
 # Colours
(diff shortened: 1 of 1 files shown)


zebster-cmd:
## Code Review

### Summary

A well-structured, readable install script with clear section headers, color-coded logging, and a logical flow. It handles the happy path well and has thoughtful touches like the `[ ! -t 0 ]` non-interactive detection for the service menu. However, there are several issues ranging from a critical `curl | bash` breakage to security concerns and cross-platform gaps that should be addressed before merging.

### Issues

#### Critical

**C1. `BASH_SOURCE[0]` is undefined when piped from `curl | bash` — local-copy path will silently misbehave**
`install.sh:112`

```bash
self="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
```

When invoked via `curl -fsSL ... | bash`, `BASH_SOURCE[0]` is empty (only set for sourced/executed files). `dirname ""` returns `.`, so `self` resolves to `$PWD`. If the user happens to have a `Makefile` and `plugin/manifest.json` in their CWD, the script will `cp -r .` their entire CWD into `~/.hermes/hermes-theia/` instead of cloning.

**Fix:**
```bash
local self="${BASH_SOURCE[0]:-}"
if [ -n "$self" ]; then
    self="$(cd "$(dirname "$self")" && pwd)"
    if [ -f "${self}/Makefile" ] && [ -f "${self}/plugin/manifest.json" ]; then
        ...
        return
    fi
fi
```

**C2. `main` runs in a subshell — `set -euo pipefail` error propagation is broken**
`install.sh:315`

```bash
if ! (main "$@"); then
```

The parentheses launch `main` in a **subshell**. `exit 1` calls inside `check_prereqs` and `setup_venv` only exit the subshell, not the script. The script continues past line 320 and prints "Installation complete!" even on failure.

**Fix:**
```bash
trap 'err "Installation did not complete."; exit 1' ERR
main "$@"
ok "Installation complete!"
```

#### Major

**M1. `setup_venv` installs `[dev]` extras — pytest/ruff/mypy should not be installed for end-users**
`install.sh:134`

```bash
"$pip" install -e ".[dev]"
```

The `[dev]` extras (`pytest`, `ruff`, `mypy`) are developer tools. Use `"$pip" install -e "."` for the install script.

**M2. `rm -rf "$target"` in `symlink_plugin` has no guard — will delete real plugin directory if it's not a symlink**
`install.sh:169`

If `$target` (`~/.hermes/plugins/theia-constellation`) is a real directory (e.g., from `make staging`), this silently deletes it.

**Fix:**
```bash
if [ -L "$target" ]; then
    rm -f "$target"
elif [ -e "$target" ]; then
    warn "Existing non-symlink plugin at ${target}, backing up..."
    mv "$target" "${target}.bak.$(date +%s)"
fi
```

**M3. No macOS compatibility — script will half-succeed and leave a broken install**
The script doesn't check the OS. systemd options 1 & 2 silently fail on macOS. Apple ships BSD make. At minimum:
```bash
if [[ "$OSTYPE" == "darwin"* ]]; then
    SKIP_SERVICE=true  # or filter systemd options
fi
```

**M4. System service file has wrong ownership/permissions**
`install.sh:268`

`mktemp` creates 0600 owned by invoking user. `sudo mv` preserves those, leaving `/etc/systemd/system/theia-watch.service` as `user:user 0600`. Conventionally should be `root:root 0644`.

**Fix:**
```bash
sudo mv "$tmpfile" /etc/systemd/system/theia-watch.service
sudo chown root:root /etc/systemd/system/theia-watch.service
sudo chmod 644 /etc/systemd/system/theia-watch.service
```

**M5. `setup_venv` unconditionally recreates the venv on re-run — not idempotent**
Every re-run re-creates the base environment. Add an early-exit:
```bash
if [ -f "${VENV_DIR}/bin/python" ] && "${VENV_DIR}/bin/python" -c "import theia_core" 2>/dev/null; then
    ok "theia-core already installed, skipping"
    return
fi
```

#### Minor

**m1.** Argument parsing silently ignores unknown flags. Add a catch-all `*) err "Unknown option: $arg"; usage; exit 1 ;;`.

**m2.** `$HELP && { usage; exit 0; }` is a fragile idiom. Use `if $HELP; then usage; exit 0; fi`.

**m3.** `git pull --ff-only` failure is only a warning — stale code may be used for the build.

**m4.** Systemd service has no `Environment=PATH=...` — if `theia_core` shells out to other tools, they may not be found.

**m5.** `npm ci` requires a committed `package-lock.json`. Verify it's tracked.

**m6.** Header comment recommends `curl ... | bash` but doesn't mention `bash -s -- --no-service` for argument passing.

#### Nit

**N1.** Logging functions use `$1` instead of `"$@"` — multi-argument calls would lose words. Quote with `"$1"` at minimum.

**N2.** `fail=1; exit 1` on line 86 — the assignment is dead code.

### Verdict: Request Changes

Two critical issues must be addressed before merge:
1. **C1** — `BASH_SOURCE[0]` empty in `curl | bash` can cause the script to copy the wrong directory
2. **C2** — subshell `main` means "Installation complete!" prints even on failure

Plus M1, M2, and M4 should be resolved. macOS gaps (M3) and idempotency (M5) could be deferred to a follow-up if documented as known limitations.

zebster-cmd:
## Force-pushed full rewrite (testing feedback)

Previously this installer ran `make build` which builds `examples/graph.json` from the fixture sessions in `examples/sessions/*.json`. After deploying, on a fresh Hermes install, the plugin would either 404 (clean machine) or — on the dev box — silently pick up the 3-node fixture graph via a hardcoded `~/projects/hermes-hackathon-seshviz/...` fallback in `graph_data.py`. Either way, the user never saw their real session data.

**This rewrite addresses all of that:**

### `install.sh` — full rewrite (no Makefile dependency)
- Build steps inlined: `npm ci` + `vite build` + manual plugin assembly via `cp`
- `pip install` without `[dev]` extras (no pytest/ruff/mypy for end-users)
- Generates `~/.hermes/theia-graph.json` from the user's **real** `~/.hermes/state.db` via `theia-core` (no fixture data ever touches the deployed plugin)
- `--dev` flag: skips panel build, plugin expects Vite at `localhost:5173`
- Default (staging): full bundled panel + pre-built graph
- `--no-graph`, `--no-service`, `--no-update` flags for non-interactive control
- Refuses to clobber a non-symlink plugin dir; backs it up to `*.bak.<ts>`
- Linux-only watcher service; macOS/non-systemd cases noted, not failed
- `trap ERR` replaces the fragile `if ! (main "$@")` subshell pattern (review issue C2)
- `BASH_SOURCE` guarded for `curl | bash` invocation (review issue C1)
- Unknown flags now error instead of being silently ignored
- System service unit installed via `install -m 0644 -o root -g root` (correct ownership/perms — review issue M4)

### `plugin/api/graph_data.py` — drop dev path leakage
- Removed the hardcoded `~/projects/hermes-hackathon-seshviz/.../examples/graph.json` fallback that leaked fixture data on the dev machine
- Honors `$THEIA_HOME` (defaults to `~/.hermes/`) for the canonical `theia-graph.json` location
- Search paths reduced to two: canonical + bundled-fallback (`<plugin>/data/graph.json`)

### Verified end-to-end on a fresh install

```text
[ OK ]  Plugin deployed: ~/.hermes/plugins/theia-constellation  (10 files)
[ OK ]  Wrote ~/.hermes/theia-graph.json
```

| Endpoint | Result |
|---|---|
| `GET /health` | `{"status":"ok","env":"staging"}` |
| `GET /graph` | **236 nodes, 55 edges**, umap projection |
| Edge kinds detected | `memory-share`, `cross-search`, `subagent`, `tool-overlap` |

(Previously this would have been the 3-node fixture or a 404.)

### Review checklist (from earlier feedback)

| Issue | Status |
|---|---|
| C1 — `BASH_SOURCE[0]` empty in `curl \| bash` | Fixed (guarded with `${BASH_SOURCE[0]:-}`) |
| C2 — subshell `main` swallows `exit 1` | Fixed (replaced with `trap ERR`) |
| M1 — `[dev]` extras installed for end-users | Fixed (plain `pip install -e`) |
| M2 — `rm -rf` on non-symlink plugin dir | Fixed (backup `.bak.<ts>` instead) |
| M3 — No macOS handling | Watcher path now Linux-only with explicit message |
| M4 — System service file owner/perms | Fixed (`sudo install -m 0644 -o root -g root`) |
| M5 — Non-idempotent venv recreate | Fixed (reuses existing venv) |
| m1 — Silently ignored unknown flags | Fixed (now errors) |
| m6 — `curl \| bash -s --` for flag passing | Documented in `--help` |

PR #15, opened 2026-04-24
feat: QoL improvements — eliminate physics jitter on filter change, search keyboard nav, loading indicator, and more

arm64be:
- Fix jittery physics readjustment when toggling edge filters by starting simulation near equilibrium (alpha=0.02) instead of full force
- Make edge pulsing phase deterministic (hash-based) so animation doesn't jump on filter toggle
- Clamp tooltip to viewport bounds to prevent overflow
- Show pointer cursor on node hover for better affordance
- Debounce search input (80ms) and add keyboard navigation (ArrowUp/Down, Enter, Escape)
- Close side panel on Escape key
- Show loading indicator during initial graph fetch and reload
- Display friendlier filter labels (Memory Share instead of MEMORY-SHARE)
- Cache JSON Schema validator with functools.cache for watch-mode reuse
- Add CLI help text for --projection, --include-features, --disable-tool-overlap
- Replace assert with proper if/raise ValueError for production safety

### DIFF
--- a/theia-core/theia_core/__main__.py
+++ b/theia-core/theia_core/__main__.py
@@ -66,9 +66,18 @@ def main(argv: list[str] | None = None) -> int:
         default=theia_home / "theia-graph.json",
         help="output graph JSON path (default: $THEIA_HOME/theia-graph.json)",
     )
-    parser.add_argument("--projection", choices=["pca", "umap", "tool-vector"], default="umap")
-    parser.add_argument("--include-features", action="store_true")
-    parser.add_argument("--disable-tool-overlap", action="store_true")
+    parser.add_argument(
+        "--projection",
+        choices=["pca", "umap", "tool-vector"],
+        default="umap",
+        help="dimensionality reduction method (default: umap)",
+    )
+    parser.add_argument(
+        "--include-features", action="store_true", help="include per-node feature vectors in output"
+    )
+    parser.add_argument(
+        "--disable-tool-overlap", action="store_true", help="skip tool-overlap edge detection"
+    )
     parser.add_argument(
         "--watch", action="store_true", help="regenerate graph when the database changes"
     )
--- a/theia-core/theia_core/__main__.py
+++ b/theia-core/theia_core/__main__.py
@@ -52,25 +52,28 @@ def _build(
 def main(argv: list[str] | None = None) -> int:
     theia_home = _theia_home()
 
-    parser = argparse.ArgumentParser(prog="theia-core")
+    parser = argparse.ArgumentParser(
+        prog="theia-core",
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter,
+    )
     parser.add_argument(
         "--db-path",
         type=Path,
         default=theia_home / "state.db",
-        help="path to Hermes SQLite database (default: $THEIA_HOME/state.db)",
+        help="path to Hermes SQLite database",
     )
     parser.add_argument(
         "-o",
         "--out",
         type=Path,
         default=theia_home / "theia-graph.json",
-        help="output graph JSON path (default: $THEIA_HOME/theia-graph.json)",
+        help="output graph JSON path",
     )
     parser.add_argument(
         "--projection",
         choices=["pca", "umap", "tool-vector"],
         default="umap",
-        help="dimensionality reduction method (default: umap)",
+        help="dimensionality reduction method",
     )
     parser.add_argument(
         "--include-features", action="store_true", help="include per-node feature vectors in output"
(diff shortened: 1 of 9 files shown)


zebster-cmd:
## Code Review

### Summary

A well-scoped QoL PR that bundles several small, user-facing polish improvements: eliminating physics jitter on filter toggle, deterministic edge pulse phases, a loading indicator, tooltip clamping, pointer cursor on hover, search debounce + keyboard navigation, and Escape-to-close on the side panel. The Python-side changes are cosmetic (argparse help strings) plus a good `assert` → `ValueError` migration and a validator cache. Overall solid work with one bug, one error-path issue, and a significant cross-PR conflict.

### Issues

#### Major

**1. Tooltip clamping ignores the CSS `transform: translate(10px, 10px)` still applied to the element**
`theia-panel/src/ui/Tooltip.ts:54-55`

The tooltip element has `transform: translate(10px, 10px)` in its base style (`Tooltip.ts:37`). The new clamping computes max positions based on `container.clientWidth - el.offsetWidth - tooltipOffset`, but the CSS transform adds an *additional* 10px shift in both axes *after* positioning. The tooltip can still overflow the container by up to 10px on the right/bottom edge.

Either remove the CSS `transform: translate(10px, 10px)` and bake the offset entirely into JS, or account for it (20px total) in the clamp. Also missing a `Math.max(0, ...)` lower-bound clamp for top-left corner.

**2. Initial `loadGraph` failure crashes the mount function with a secondary error**
`theia-panel/src/index.ts:256-267`

The `try/finally` correctly hides the loading overlay, but if `loadGraph` throws, execution continues to `setupGraph(initialGraph)` — where `initialGraph` is uninitialized (`let` but never assigned). This throws a `ReferenceError`/use-before-assign, masking the original error.

**Fix:**
```ts
try {
  initialGraph = await loadGraph(graphUrl);
} catch (err) {
  loadingEl.style.opacity = "0";
  throw err;
}
loadingEl.style.opacity = "0";
setTimeout(() => { ... }, 300);
```

**3. PR #13 (`feat/more-filters`) will conflict heavily with this PR**

Both PRs independently rewrite `SearchBar.ts` with keyboard navigation (`selectedIndex`, `currentResults`, `ArrowUp/Down/Enter/Escape`). PR #13 also changes the `createFilterBar` signature (adds `graph` param and `FilterState` return type). Merging both will require manual resolution of essentially the entire `SearchBar.ts`, plus `FilterBar.ts` and `index.ts`. Pick a merge order and rebase the second one.

#### Minor

**4. Loading indicator code is duplicated between initial load and `reload()`**
The initial load fades out gracefully; `reload()` removes the overlay instantly causing a visual flash. Extract to a `showOverlay(text) -> remove()` helper and use the same fade transition for both.

**5. `@functools.cache` on `_load_validator()` permanently caches a file read**
`theia-core/theia_core/emit.py:43-45` — Unbounded permanent memo. In `--watch` mode, a schema update mid-session would be silently ignored. Consider documenting the assumption or using `lru_cache(maxsize=1)`.

**6. `hash01` has limited precision — only 1000 discrete values**
`(fnv1a(str) % 1000) / 1000` → ~12% chance of at least one collision at 50 edges (birthday paradox). Using `fnv1a(str) / 0xFFFFFFFF` would give ~4 billion discrete values at zero extra cost.

**7. SearchBar: `Escape` does nothing when `currentResults.length === 0`**
The keydown handler returns early, so pressing Escape with typed text (before debounce fires) doesn't blur or clear. Tight race but possible.

**8. SidePanel `outline:none` + `tabIndex=-1` may affect keyboard-only users**
Removes the focus indicator entirely. For a data-viz panel this is acceptable, but a custom `:focus-visible` style (e.g., border color change) would be a low-effort accessibility win.

#### Nit

**9.** `letter-spacing` inconsistency between filter bar (`0.05em`) and loading overlay (`0.1em`).

**10.** `kindLabels` typed as `Record<string, string>` won't catch typos. Use `satisfies Record<TheiaGraph["edges"][number]["kind"], string>`.

**11.** Argparse help text says `(default: umap)` redundantly — would be auto-appended with `ArgumentDefaultsHelpFormatter`. Minor consistency point.

**12.** Empty catch on `element.removeChild(loadingEl)` — a comment explaining why removal can fail (mount destroyed during load) would help maintainers.

### Verdict: Request Changes

The tooltip clamping bug (Major #1) will cause visible overflow in production. The `loadGraph` failure path (Major #2) produces confusing double errors. Both are small fixes. The PR #13 conflict (Major #3) is a process concern but important to surface now. After fixing #1 and #2, this is a clean approve.

PR #16, opened 2026-04-24
fix(install): rewrite installer — real-data graph, staging by default

zebster-cmd:
# fix(install): rewrite installer — real-data graph, staging by default

Follow-up fix to PR #14 (already merged). Testing on a fresh Hermes install surfaced two bugs that combine to make the deployed plugin serve fixture data instead of the user's real sessions.

## The bug

When `install.sh` ran on a fresh machine, the dashboard's Constellation tab showed the **3-node fixture graph** (built from `examples/sessions/*.json`) instead of the user's actual Hermes sessions.

Two causes:

1. **`install.sh` ran `make build`**, whose `build-graph` target seeds a test DB from `examples/sessions/*.json` fixtures and writes a fixture graph — never touches the user's real `~/.hermes/state.db`.
2. **`plugin/api/graph_data.py` had a hardcoded dev-machine fallback** (`~/projects/hermes-hackathon-seshviz/.../examples/graph.json`) that masked the install bug on the original dev box, since `~/.hermes/theia-graph.json` happened to exist there from earlier work.

### `install.sh` — full rewrite

Drops the Makefile dependency entirely. Build steps are inlined (`npm ci` + `vite build` + manual plugin assembly via `cp`). The graph is now generated by invoking `theia-core` directly against the user's real `~/.hermes/state.db`:

```bash
"${VENV_DIR}/bin/theia-core" --db-path "$STATE_DB" -o "$GRAPH_OUT"
```

Other improvements addressing the review of the original PR #14:

| Issue | Status |
|---|---|
| C1 — `BASH_SOURCE[0]` empty in `curl \| bash` | Fixed (guarded with `${BASH_SOURCE[0]:-}`) |
| C2 — subshell `main` swallowed `exit 1` | Fixed (replaced with `trap ERR`) |
| M1 — `[dev]` extras installed for end-users | Fixed (plain `pip install -e`) |
| M2 — `rm -rf` on non-symlink plugin dir | Fixed (backs up to `*.bak.<ts>`) |
| M3 — No macOS handling | Watcher path now Linux-only with explicit message |
| M4 — System service file owner/perms | Fixed (`sudo install -m 0644 -o root -g root`) |
| M5 — Non-idempotent venv recreate | Fixed (reuses existing venv) |
| m1 — Silently ignored unknown flags | Fixed (now errors) |
| m6 — `curl \| bash -s --` for flag passing | Documented in `--help` |

New flags:
- `--dev` — skip panel build, plugin expects Vite at `localhost:5173`
- `--no-graph` — skip initial graph generation
- `--no-service` — skip watcher prompt
- `--no-update` — don't `git pull` an existing clone

### `plugin/api/graph_data.py` — drop dev path leakage

- Removed the hardcoded `~/projects/hermes-hackathon-seshviz/.../examples/graph.json` fallback
- Honors `$THEIA_HOME` (defaults to `~/.hermes/`) for the canonical `theia-graph.json` location
- Search paths reduced to two: `$THEIA_HOME/theia-graph.json` (canonical) + `<plugin>/data/graph.json` (bundled fallback)

### Makefile — kept

The Makefile stays as a developer/CI tool. It does file-based incremental rebuilds, golden-fixture regen (`make build-graph`), and versioned packaging (`make package`) that `release.yml` depends on. The original problem wasn't "the Makefile exists" — it was "install.sh shouldn't depend on `make build` because that builds from fixtures". That's now fixed without touching the Makefile.

## Verified end-to-end on a fresh install

```text
[INFO]  Generating initial graph from /home/.../.hermes/state.db...
wrote /home/.../.hermes/theia-graph.json — 236 nodes, 55 edges
[ OK ]  Plugin deployed: ~/.hermes/plugins/theia-constellation  (10 files)
```

| Endpoint | Result |
|---|---|
| `GET /health` | `{"status":"ok","env":"staging"}` |
| `GET /graph` | **236 nodes, 55 edges**, umap projection (was 3-node fixture) |
| Edge kinds detected | `memory-share`, `cross-search`, `subagent`, `tool-overlap` |

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,23 +1,28 @@
 #!/usr/bin/env bash
 # ---------------------------------------------------------------------------
-# theia -- One-command installer
+# Theia Constellation — One-command installer
+#
+# Builds and deploys the Theia plugin into a Hermes Agent installation.
+# Default target is staging (production-style: bundled panel, pre-built graph
+# from the real ~/.hermes/state.db).  Pass --dev to leave the panel pointing
+# at a Vite dev server instead.
 #
 # Usage:
-#   bash <(curl -fsSL https://raw.githubusercontent.com/arm64be/theia/main/install.sh)
-#   bash <(curl -fsSL https://raw.githubusercontent.com/arm64be/theia/main/install.sh) -s -- --no-service
-#   bash install.sh                            (from a local clone)
-#   bash install.sh --help                     (show this help)
-#   bash install.sh --no-service               (skip service prompt)
-#   bash install.sh --no-update                (don't git pull if already installed)
+#   bash install.sh                          install staging build
+#   bash install.sh --dev                    dev mode (panel via Vite)
+#   bash install.sh --no-graph               skip initial graph generation
+#   bash install.sh --no-service             skip the watcher service prompt
+#   bash install.sh --no-update              don't git-pull an existing clone
+#   bash install.sh --help
 #
-# What it does:
-#   1. Checks prerequisites (Python >= 3.11, Node.js, npm, GNU Make)
-#   2. Clones the repo into ~/.hermes/hermes-theia/ (git pull to update if exists)
-#   3. Creates a Python virtual environment and installs dependencies
-#   4. Installs Node dependencies
-#   5. Builds the panel embed and assembles the plugin
-#   6. Symlinks the plugin into ~/.hermes/plugins/theia-constellation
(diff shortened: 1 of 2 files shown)


zebster-cmd:
Regression filed: #66 — install.sh copies fresh plugin_api.py from repo, overwriting local relative-import workarounds and re-introducing the  crash.

PR #22, opened 2026-04-25
fix(panel): only click the currently-hovered node (#20)

### LINKED ISSUE #20: Inconsistent node clicking: side panel fails to open when mouse leaves iframe before mouseup
## Problem

Clicking on constellation nodes is inconsistent — sometimes the side panel pops up with session details, sometimes nothing happens at all. This is especially noticeable when clicking quickly or with slight mouse drift.

## Reproduction

1. Open the Constellation tab in the Hermes dashboard.
2. Click a node with a quick "tap" motion (slight mouse movement during press/release).
3. About 30–50% of the time, the side panel does not open.
4. If you hold the mouse perfectly still during click, it works reliably.

## Root cause

The click/drag event handling in `theia-panel/src/index.ts` attaches `mousemove` and `mouseup` listeners to the `element` (the container div inside the iframe). However, `mouseup` fires on the element where the mouse is **released**, not where `mousedown` occurred.

When the user clicks a node and the mouse drifts outside the iframe boundary before release, the `mouseup` event fires on the **parent document** instead of the iframe's `element`. The `element`-scoped listener is never triggered. This means:

1. `isMouseDown` leaks to `true` (never cleared), corrupting the next interaction.
2. The click logic in the `mouseup` handler is skipped entirely.

Additionally, the `pickAt()` call in the click handler uses a tight `radiusScale = 0.35`, which already makes edge clicks sensitive — combining this with the missed `mouseup` makes the issue feel frequent and unpredictable.

### Relevant code

`theia-panel/src/index.ts` (lines 340–381):

```typescript
element.addEventListener("mousedown", (e) => {
  if (e.button === 0) dragMode = "rotate";
  else if (e.button === 2) dragMode = "pan";
  else return;
  isMouseDown = true;
  hasDragged = false;
  mouseDownPos = { x: e.clientX, y: e.clientY };
});

element.addEventListener("mousemove", (e) => { ... });

element.addEventListener("mouseup", (e) => {
  // This is never called if mouseup happens outside the iframe!
  if (isMouseDown && !hasDragged && performance.now() - lastWheelAt >= 200) {
    const idx = picker.pickAt(e.clientX, e.clientY, 0.35);
    if (idx !== null) {
      sidePanel.show(n, related);
      emit("node-click", n.id);
    }
  }
  isMouseDown = false;
  hasDragged = false;
  dragMode = null;
});
```

## Expected behavior

Node clicks should be reliable regardless of whether the mouse leaves the iframe boundary before release. The click target should register as long as the pointer was captured during `mousedown`.

### Option A: Pointer Events with `setPointerCapture` (recommended)

Replace mouse events with Pointer Events and use `setPointerCapture` to guarantee `pointerup` fires even when the pointer leaves the element:

```typescript
element.addEventListener("pointerdown", (e) => {
  if (e.button === 0) dragMode = "rotate";
  else if (e.button === 2) dragMode = "pan";
  else return;
  (e.target as HTMLElement).setPointerCapture(e.pointerId);
  isMouseDown = true;
  hasDragged = false;
  mouseDownPos = { x: e.clientX, y: e.clientY };
});

element.addEventListener("pointermove", (e) => { ... });

element.addEventListener("pointerup", (e) => {
  // Fires reliably because of setPointerCapture
  if (isMouseDown && !hasDragged && performance.now() - lastWheelAt >= 200) {
    const idx = picker.pickAt(e.clientX, e.clientY, 0.35);
    if (idx !== null) {
      sidePanel.show(n, related);
      emit("node-click", n.id);
    }
  }
  isMouseDown = false;
  hasDragged = false;
  dragMode = null;
});
```

### Option B: Window-level listeners during drag

Keep `mousedown` on `element`, but attach `mousemove`/`mouseup` to `window` during active drag, and remove them on release. This is the classic drag-pattern fix.

## Environment

- Hermes Agent: latest main
- Theia: main branch
- Browser: Chrome / Firefox (issue is iframe-agnostic)
- OS: WSL2 (Windows host accessing dashboard via portproxy)

---

**Workaround for users:** Click nodes slowly and hold the mouse perfectly still until releasing. Avoid quick taps or dragging motions.

zebster-cmd:
Closes #20.

## Root cause

`theia-panel/src/index.ts` attached `mousedown`/`mousemove`/`mouseup` directly to the canvas `element`. `mouseup` fires on the element where the pointer is *released*, not where `mousedown` occurred — so a quick tap with even a few pixels of mouse drift would route the up event to the parent document instead of the iframe. Two consequences:

1. `isMouseDown` was stuck at `true`, corrupting the next interaction.
2. The click branch inside the up-handler was skipped, so the side panel never opened.

The reporter measured a ~30–50% failure rate on quick taps; reproducing the same DOM topology in this repo confirms the diagnosis.

## Fix

Adopt **Option A** from the issue write-up: Pointer Events with `setPointerCapture`. After `pointerdown` captures the pointer, all subsequent `pointermove`/`pointerup` events are routed to `element` regardless of where the pointer goes — including outside the iframe.

Specifics in `theia-panel/src/index.ts:329-419`:

- `mousedown` → `pointerdown` + `setPointerCapture(e.pointerId)`.
- `mousemove` (drag) → `pointermove`.
- `mouseup` → `pointerup`, plus a new `pointercancel` handler that mirrors the cleanup path so an interrupted gesture (e.g. system-level scroll cancellation on touch) doesn't leave `isMouseDown` stuck.
- The hover-tracking `mousemove` for `lastMouse` (used by the tooltip) is converted to `pointermove` for consistency.
- Cleanup is centralised in `releasePointer()` which releases the captured pointer and resets `isMouseDown` / `hasDragged` / `dragMode`. `setPointerCapture` / `releasePointerCapture` are wrapped in try/catch because either can throw if the pointer became inactive between calls.

### Behaviour preserved

- Same drag threshold (3px) before a press is treated as a drag.
- Same button mapping (left = rotate, right = pan).
- Same 200ms suppression window after a wheel event before a click registers.
- Same picker radius (`0.35`) on click.

## Test plan

Full panel CI run locally — every gate from `.github/workflows/panel.yml`:

```
npm ci
npm run generate-types
npm run typecheck
npm run test          # 2/2 passing (data/load.test.ts)
npm run build         # 24 modules, 70.67 kB / 19.45 kB gzip
npm run format:check  # All matched files use Prettier code style!
```

End-to-end smoke check via `bash install.sh --no-service --no-graph` — plugin rebuilds, deploys to `~/.hermes/plugins/theia-constellation`, dashboard ready to load.

(`schemas/graph.schema.json` was unchanged, so `npm run generate-types` produced an unrelated formatting tweak in `theia-panel/src/data/types.ts` from a json2ts version drift; reverted to keep this PR scoped to the click fix. CI regenerates the file before typecheck regardless.)

## Manual repro to verify the fix

1. `bash install.sh --no-service --no-graph`
2. Open the Constellation tab in the Hermes dashboard.
3. Tap nodes quickly with a few px of mouse drift. Before this PR: ~30–50% silent failures. After: every tap opens the side panel.

### DIFF
--- a/theia-panel/src/index.ts
+++ b/theia-panel/src/index.ts
@@ -363,9 +363,19 @@ export async function mount(
     }
   });
 
-  element.addEventListener("mouseup", (e) => {
+  element.addEventListener("mouseup", () => {
     if (isMouseDown && !hasDragged && performance.now() - lastWheelAt >= 200) {
-      const idx = picker.pickAt(e.clientX, e.clientY, 0.35);
+      // Click target is whatever is currently hovered — and only that.
+      // The hover picker (Picker.onMove, radius 1.0) drives the tooltip
+      // and pointer cursor, so the user already has a clear visual cue
+      // for what they are about to click. A fresh pickAt here at any
+      // radius can disagree with the hover state — different radius,
+      // fresh hit test against a still-settling simulation — producing
+      // the "tooltip says hovered but click does nothing" symptom on
+      // first load (#20), or worse, a neighbour the user wasn't aiming
+      // at. If nothing is hovered, the click is a no-op rather than a
+      // positional guess.
+      const idx = picker.currentHovered();
       if (idx !== null) {
         const n = currentGraph.nodes[idx]!;
         const related = currentGraph.edges.filter(


zebster-cmd:
Closing — superseded by #23. Mel's one-line widening of the click radius is the simpler fix and lands the same outcome empirically (the hover/click radius mismatch in `Picker.pickAt`); the additional pointer-events migration this PR carried turned out not to change behaviour in practice.

zebster-cmd:
Fixes #20. A cleaner one-line alternative is being explored in PR #23.

PR #30, opened 2026-04-25
fix(panel): responsive filter/search/panel layout (#27)

### LINKED ISSUE #27: bug(panel): search bar, filter bar, and side panel overlap on narrow viewports
## Bug Report

|---|---|---|

### Problems observed

### Repro

4. Observe:

### Suggested fixes

### Environment

zebster-cmd:
Closes #27.

## Problem

The filter bar, search bar, and side panel were each independently absolute-positioned with hard-coded widths that did not account for one another. As reported in #27:

1. On viewports ~800–1000px the centered search bar (50vw) reached back over the filter bar (top-left, no wrap).
2. With the side panel open the search bar (`z-index: 10`) rendered on top of the panel; its dropdown opened down into the panel area.
3. The filter bar overflowed its own padded background because the flex row had no `flex-wrap`.
4. At 800px the side panel (45vw = 360px) plus the filter bar squeezed the graph into ~190px.

## Approach

Centralise all positional / responsive rules in a new `theia-panel/src/ui/layoutStyles.ts` that injects a single idempotent `<style>` element. Each component keeps **only** theme-derived properties inline (background, border colour, text colour) and adds a class for everything else. `SidePanel` toggles `.tp-panel-open` on its container so `SearchBar` reacts via pure CSS — no `ResizeObserver`, no inter-component callbacks.

Why this is the cleanest option for maintenance:

- Zero new dependencies; one ~85-line module.
- Single place to read and edit the rules when the layout evolves.
- Theme/colour stays where it lives now (component-level), so theme switching does not have to touch the layout sheet.
- Native CSS `@media` query handles the narrow-viewport behaviour — no JS resize listeners.
- Class-driven transform/transition replaces the previously fragile pattern of restating `transform` inside `cssText` derived from `currentId`.

## Behaviour after fix

| Viewport / state | Filter bar | Search bar | Side panel |
|---|---|---|---|
| Wide, panel closed | top-left, wraps if needed | top-right, `clamp(180,30vw,320)` | hidden (slides off-right) |
| Wide, panel open | top-left, wraps if needed | shifted left of panel: `right: panelWidth + 12px` | `min(420px, 45vw)` |
| ≤900px, panel closed | top-left, wraps | top-right (clamped) | hidden |
| ≤900px, panel open | top-left, wraps | bottom-left of canvas, dropdown opens **upward** | `min(320px, 85vw)` |

## Files

- `theia-panel/src/ui/layoutStyles.ts` — new, single style injection.
- `theia-panel/src/ui/FilterBar.ts` — added `tp-filter-bar` class; cssText now sets only `background`, `border`, `color`, `font`, etc.
- `theia-panel/src/ui/SearchBar.ts` — added `tp-search-bar` and `tp-search-dropdown` classes; preserved dynamic `display` state across theme updates.
- `theia-panel/src/ui/SidePanel.ts` — added `tp-side-panel` class; `show()`/`hide()` toggle `.tp-side-panel--open` on the panel and `.tp-panel-open` on the container.

## Verification

```
$ npm run typecheck   # tsc --noEmit  → clean
$ npm run build       # vite build    → 72.52 kB │ gzip: 20.03 kB
$ npm test            # vitest run    → 2/2 passing
$ npm run format:check # prettier     → clean
```

Manual repro (per issue reproduction steps): open the Constellation tab, resize to ~800px, click a node, and observe the search bar now sits at the bottom-left clear of both panel and filter bar; toggles wrap onto a second row instead of overflowing.

### DIFF
--- a/theia-panel/src/ui/SearchBar.ts
+++ b/theia-panel/src/ui/SearchBar.ts
@@ -1,4 +1,5 @@
 import type { TheiaGraph } from "../data/types";
+import { ensureLayoutStyles } from "./layoutStyles";
 import type { ThemeTokens } from "./Theme";
 import { themeBgAlpha } from "./Theme";
 import { escape } from "./utils";
@@ -15,21 +16,25 @@ export function createSearchBar(
   initialTheme: ThemeTokens,
   isVisible?: (node: TheiaGraph["nodes"][number]) => boolean,
 ) {
+  ensureLayoutStyles();
+
   let theme = initialTheme;
 
   const wrapper = document.createElement("div");
+  wrapper.classList.add("tp-search-bar");
   const input = document.createElement("input");
   const dropdown = document.createElement("div");
+  dropdown.classList.add("tp-search-dropdown");
+  dropdown.style.display = "none";
 
   let currentResults: SearchResult[] = [];
   let selectedIndex = -1;
 
+  // Position, width, font and dropdown placement live in .tp-search-bar /
+  // .tp-search-dropdown (see layoutStyles.ts). Inline styles cover only
+  // theme-derived properties so theme changes don't disturb the layout.
   function applyWrapperStyle() {
-    wrapper.style.cssText = `
-      position: absolute; top: 12px; left: 50%; transform: translateX(-50%);
-      z-index: 10; font: 13px/1.4 'Mondwest', var(--theia-font, ui-monospace, monospace);
-      color: #${theme.fg}; width: min(320px, 50vw);
-    `;
+    wrapper.style.color = `#${theme.fg}`;
   }
 
   function applyInputStyle() {
@@ -43,11 +48,12 @@ export function createSearchBar(
   }
 
   function applyDropdownStyle() {
+    // Preserve dynamic display state across theme updates.
+    const display = dropdown.style.display;
     dropdown.style.cssText = `
-      position: absolute; top: calc(100% + 6px); left: 0; right: 0;
       background: ${themeBgAlpha(theme, 0.95)}; border: 1px solid #${theme.border};
-      display: none;
-      backdrop-filter: blur(6px); max-height: 240px; overflow-y: auto;
+      backdrop-filter: blur(6px);
+      display: ${display || "none"};
     `;
   }
 
--- a/theia-panel/src/ui/SearchBar.ts
+++ b/theia-panel/src/ui/SearchBar.ts
@@ -39,8 +39,7 @@ export function createSearchBar(
 
   function applyInputStyle() {
     input.style.cssText = `
-      width: 100%; box-sizing: border-box;
-      padding: 8px 12px; background: ${themeBgAlpha(theme, 0.85)};
+      background: ${themeBgAlpha(theme, 0.85)};
       border: 1px solid #${theme.border};
       color: #${theme.fg}; font: inherit; outline: none;
       backdrop-filter: blur(4px);
--- a/theia-panel/src/ui/SidePanel.ts
+++ b/theia-panel/src/ui/SidePanel.ts
@@ -1,4 +1,5 @@
 import type { TheiaGraph } from "../data/types";
+import { ensureLayoutStyles } from "./layoutStyles";
 import type { ThemeTokens } from "./Theme";
 import { themeBgAlpha } from "./Theme";
 import { escape, truncate } from "./utils";
@@ -48,18 +49,22 @@ export function createSidePanel(
   initialTheme: ThemeTokens,
   onNavigate?: (nodeId: string) => void,
 ) {
+  ensureLayoutStyles();
+
   let theme = initialTheme;
   const el = document.createElement("aside");
+  el.classList.add("tp-side-panel");
   let currentId: string | null = null;
 
+  // Position, sizing, padding, transform/transition live in .tp-side-panel
+  // (see layoutStyles.ts). Inline styles cover only theme-derived
+  // properties; show()/hide() toggle the .tp-side-panel--open class on this
+  // element and the .tp-panel-open class on the container so SearchBar can
+  // react to the open state via pure CSS.
   function applyPanelStyle() {
     el.style.cssText = `
-      position: absolute; top: 0; right: 0; bottom: 0; width: min(420px, 45vw);
       background: ${themeBgAlpha(theme, 0.95)}; border-left: 1px solid #${theme.border};
       color: #${theme.fg}; font: 13px/1.6 var(--theia-font, ui-monospace, monospace);
-      transform: translateX(${currentId ? "0" : "100%"}); transition: transform 220ms ease-out;
-      padding: 20px 22px; overflow-y: auto; overscroll-behavior: contain;
-      box-sizing: border-box; outline: none;
     `;
   }
   applyPanelStyle();
@@ -87,7 +92,8 @@ export function createSidePanel(
     lastNode = node;
     lastEdges = relatedEdges;
     renderContent();
-    el.style.transform = "translateX(0)";
+    el.classList.add("tp-side-panel--open");
+    container.classList.add("tp-panel-open");
     el.focus({ preventScroll: true });
     scrollTop();
   }
@@ -173,7 +179,8 @@ export function createSidePanel(
 
   function hide() {
     currentId = null;
-    el.style.transform = "translateX(100%)";
+    el.classList.remove("tp-side-panel--open");
+    container.classList.remove("tp-panel-open");
   }
 
   function currentNodeId() {
@@ -187,6 +194,7 @@ export function createSidePanel(
     }
   }
   function dispose() {
+    container.classList.remove("tp-panel-open");
     container.removeChild(el);
   }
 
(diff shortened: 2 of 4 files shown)


zebster-cmd:
Fixes #27.

theia-panel/src/ui/layoutStyles.ts
```diff
@@ -0,0 +1,88 @@
...
+  if (injected) return;
+  injected = true;
+
+  const style = document.createElement("style");
+  style.dataset.theia = "panel-layout";
+  style.textContent = `
+    /* Filter bar — top-left. Wraps to multiple rows when toggles + model
+       dropdown can't fit, so it never overflows its background. min-height
+       matches the search bar so both align on the top row. */
+    .tp-filter-bar {
```
zebster-cmd, line 27: **WARNING**: SearchBar is repositioned from horizontal-center (`left: 50%; transform: translateX(-50%)`) to top-right (`right: 12px`). This is a visible UX change presented inside what reads as a refactor — worth calling out explicitly in the PR description, and double-checking it's the intended visual outcome rather than an accidental side-effect of the layout consolidation.

💡 If centering was intentional, restore via `left: 50%; transform: translateX(-50%);` and adjust the `.tp-panel-open` rule to shift left from center instead of changing `right`. If the new top-right alignment is intended, mention the visual change in the commit/PR copy.

theia-panel/src/ui/layoutStyles.ts
```diff
@@ -0,0 +1,88 @@
...
+ *
+ * Theme-dependent properties (colors, fonts) stay inline on each element so
+ * theme switching does not have to touch this stylesheet.
+ */
+
+let injected = false;
+
+export function ensureLayoutStyles(): void {
+  if (injected) return;
+  injected = true;
```
zebster-cmd, line 19: **WARNING**: FilterBar uses `max-width: calc(100% - 24px)` and `flex-wrap: wrap`, but the new SearchBar is right-anchored (`right: 12px`) at the same `top: 12px`. Between ~900px (where the SearchBar gets `display:none`) and ~1100–1200px (where toggles + model select fit comfortably with room for the search bar), a fully populated FilterBar can grow into the SearchBar's footprint and visually collide. The old FilterBar didn't reserve right-side space either, but the SearchBar used to be centered with `width: min(320px, 50vw)`, so the collision zone was different.

💡 Cap FilterBar width to the space left of the SearchBar, e.g. `max-width: calc(100% - 24px - clamp(180px, 30vw, 320px) - 12px);` (and drop that reservation under the `@media (max-width: 900px)` rule where the SearchBar is hidden). Or move the SearchBar to a different row when both are present.

theia-panel/src/ui/layoutStyles.ts
```diff
@@ -0,0 +1,88 @@
...
+ * one place to read/edit when the layout evolves.
+ *
+ * Theme-dependent properties (colors, fonts) stay inline on each element so
+ * theme switching does not have to touch this stylesheet.
+ */
+
+let injected = false;
+
+export function ensureLayoutStyles(): void {
+  if (injected) return;
```
zebster-cmd, line 18: **SUGGESTION**: Old FilterBar had `padding: 6px 14px` (top/bottom breathing room). New rule is `padding: 0 14px` with `min-height: 32px`, which works for one row but, when the bar wraps, items in subsequent rows butt directly against the top/bottom borders. `gap: 14px` provides row spacing between wrapped rows but no padding above the first or below the last row.

💡 Restore symmetric vertical padding, e.g. `padding: 6px 14px;` and drop `min-height: 32px` (the inner toggle height + padding will set bar height naturally), or keep `min-height` and add `padding-block: 6px;`.

theia-panel/src/ui/layoutStyles.ts
```diff
@@ -0,0 +1,88 @@
...
+      max-height: 240px; overflow-y: auto;
+    }
+
+    /* Side panel — slides in from the right. */
+    .tp-side-panel {
+      position: absolute; top: 0; right: 0; bottom: 0;
+      width: min(420px, 45vw);
+      padding: 20px 22px;
+      overflow-y: auto; overscroll-behavior: contain;
+      box-sizing: border-box; outline: none;
```
zebster-cmd, line 64: **SUGGESTION**: `min(420px, 45vw)` is duplicated between `.tp-side-panel { width: ... }` (line 49) and `.tp-panel-open .tp-search-bar { right: calc(... + 12px) }` (line 64). If the panel width ever changes, these will silently drift out of sync and the SearchBar will tuck under the panel.

💡 Extract to a custom property at a common ancestor (or `:root`): `--tp-side-panel-width: min(420px, 45vw);`, then reference it in both rules. The `@media (max-width: 900px)` block can override the variable too.

theia-panel/src/ui/layoutStyles.ts
```diff
@@ -0,0 +1,88 @@
...
+ * viewport. We do that with plain CSS classes + media queries instead of
+ * bespoke JS resize/coordination logic — no dependencies, no listeners, and
+ * one place to read/edit when the layout evolves.
+ *
+ * Theme-dependent properties (colors, fonts) stay inline on each element so
+ * theme switching does not have to touch this stylesheet.
+ */
+
+let injected = false;
+
```
zebster-cmd, line 16: **SUGGESTION**: Module-level `let injected = false` is never reset, and the appended `<style data-theia="panel-layout">` is never removed. For the running app this is fine (singleton intent), but in tests / hot-reload / Storybook the module instance can be re-evaluated independently of the DOM, so a stale style element can be left behind without `injected` being aware, or vice versa.

💡 Make the check DOM-authoritative: `if (document.head.querySelector('style[data-theia="panel-layout"]')) return;` instead of relying on the module-level flag. Cheap and self-healing across reloads.

theia-panel/src/ui/SearchBar.ts
```diff
@@ -15,39 +16,43 @@ export function createSearchBar(
...
 
   function applyDropdownStyle() {
+    // Preserve dynamic display state across theme updates.
+    const display = dropdown.style.display;
     dropdown.style.cssText = `
-      position: absolute; top: calc(100% + 6px); left: 0; right: 0;
       background: ${themeBgAlpha(theme, 0.95)}; border: 1px solid #${theme.border};
-      display: none;
-      backdrop-filter: blur(6px); max-height: 240px; overflow-y: auto;
+      backdrop-filter: blur(6px);
```
zebster-cmd, line 54: **SUGGESTION**: Dropdown visibility is still controlled via inline `style.display` while the side panel switched to a `--open` class. `applyDropdownStyle` now has to read the current display value before reassigning `cssText` to avoid wiping it — that round-trip is fragile (e.g., if a future change sets display via a class, the read returns empty and the dropdown silently collapses).

💡 Mirror SidePanel: add a `.tp-search-dropdown--open { display: block; }` rule and toggle the class in `render`/`hide`/`commit` instead of writing to `dropdown.style.display`. Then `applyDropdownStyle` only needs theme-derived inline properties and doesn't have to preserve state.

theia-panel/src/ui/SidePanel.ts
```diff
@@ -48,18 +49,22 @@ export function createSidePanel(
...
     el.style.cssText = `
-      position: absolute; top: 0; right: 0; bottom: 0; width: min(420px, 45vw);
       background: ${themeBgAlpha(theme, 0.95)}; border-left: 1px solid #${theme.border};
       color: #${theme.fg}; font: 13px/1.6 var(--theia-font, ui-monospace, monospace);
-      transform: translateX(${currentId ? "0" : "100%"}); transition: transform 220ms ease-out;
-      padding: 20px 22px; overflow-y: auto; overscroll-behavior: contain;
-      box-sizing: border-box; outline: none;
     `;
   }
   applyPanelStyle();
```
zebster-cmd, line 70: **SUGGESTION**: Pre-existing but worth noting alongside this refactor: the focus-style sheet uses the bare selector `aside:focus-visible`, which leaks to any other `<aside>` on the host page (theia-panel is embedded in a larger Hermes dashboard). Now that you're already injecting a scoped `<style data-theia="panel-layout">`, this is a good moment to scope this one too.

💡 Replace the selector with `.tp-side-panel:focus-visible { outline: 1px solid #${theme.accent}; }` and consider folding it into `layoutStyles.ts` (with the accent color set via a CSS variable updated on theme change), so the dynamically-rebuilt `<style>` element is no longer needed.

zebster-cmd (commented):
## Code Review

Solid refactor that consolidates absolute-position/responsive layout into a single injected stylesheet, leaving inline styles only for theme-derived properties. The class/CSS-driven side-panel open state is a clear improvement over coordinating via inline `transform`. A few concerns around the new top-right SearchBar position colliding with a wrapped FilterBar between ~900px and the breakpoint, lost vertical padding on the wrapped FilterBar, and a duplicated `min(420px, 45vw)` magic number across two rules.

zebster-cmd:
Follow-up issue filed: #57 — the absolute overlay approach conflicts with dashboard viewport conventions. Screenshots in workspace/theia-ui/ show the layout problems.

arm64be:
fixed by #67

PR #33, opened 2026-04-25
fix(panel): add missing ditherTarget.setSize in Post.resize() — fixes…

### LINKED ISSUE #31: bug(panel): constellation does not resize correctly when browser window changes size
## Bug Report

When the browser window is resized, the constellation visualization does not properly update its render resolution. The canvas appears to stretch, crop, or retain the old aspect ratio.

### Root cause analysis

The resize pipeline has a missing step in the post-processing chain.

**Current resize flow (`Scene.ts` → `index.ts` → `Post.ts`):**

1. `Scene.ts` `ResizeObserver` detects container size change.
2. `Scene.resize()` updates:
   - `camera.aspect` and `camera.updateProjectionMatrix()`
   - `renderer.setSize(w, h, false)`
3. `index.ts` overrides `ctx.resize` to also call `post.resize()`.
4. `Post.resize()` updates:
   - `edgesTarget.setSize(w * dpr, h * dpr)` ✅
   - `preBloomTarget.setSize(w * dpr, h * dpr)` ✅
   - `sceneTarget.setSize(w * dpr, h * dpr)` ✅
   - `composer.setSize(w, h)` ✅
   - `bloom.setSize(w, h)` ✅
   - **`ditherTarget.setSize(...)` ❌ MISSING**

**Why this breaks the render:**

The dithering pass (`Post.ts:269–277`) renders the scene through a Bayer dither shader into `ditherTarget`, then the overlay pass composites `ditherTarget.texture` + `sceneTarget.texture`. If `ditherTarget` stays at its initial dimensions while everything else scales, the composited output is sampled at the wrong resolution — causing the visual to appear stretched, cropped, or pixelated after resize.

### Repro

2. Resize the browser window (make it significantly narrower or wider).
3. Observe that the constellation rendering does not match the new viewport — the image appears to maintain the old aspect ratio or shows black bars / cropping.

### Suggested fix

In `theia-panel/src/scene/Post.ts`, add `ditherTarget.setSize(w2 * dpr2, h2 * dpr2)` to the `resize()` function:

```ts
function resize() {
  const w2 = container.clientWidth;
  const h2 = container.clientHeight;
  const dpr2 = renderer.getPixelRatio();
  edgesTarget.setSize(w2 * dpr2, h2 * dpr2);
  preBloomTarget.setSize(w2 * dpr2, h2 * dpr2);
  sceneTarget.setSize(w2 * dpr2, h2 * dpr2);
  ditherTarget.setSize(w2 * dpr2, h2 * dpr2); // ← add this
  composer.setSize(w2, h2);
  bloom.setSize(w2, h2);
}
```

### Environment

- File: `theia-panel/src/scene/Post.ts`

arm64be:
Closes #31.

### DIFF
--- a/theia-panel/src/scene/Post.ts
+++ b/theia-panel/src/scene/Post.ts
@@ -298,6 +298,7 @@ export function createPost(
     edgesTarget.setSize(w2 * dpr2, h2 * dpr2);
     preBloomTarget.setSize(w2 * dpr2, h2 * dpr2);
     sceneTarget.setSize(w2 * dpr2, h2 * dpr2);
+    ditherTarget.setSize(w2 * dpr2, h2 * dpr2);
     composer.setSize(w2, h2);
     bloom.setSize(w2, h2);
   }


zebster-cmd (commented):
# Code Review — PR #33

## Summary
Adds a single missing call `ditherTarget.setSize(w2 * dpr2, h2 * dpr2)` in `Post.resize()` at `theia-panel/src/scene/Post.ts:301`, so the dither render target tracks viewport changes alongside the other post-processing targets.

## Correctness
The fix is correct and precisely matches the root cause described in issue #31.

- `ditherTarget` is created with `w * dpr` dimensions at `theia-panel/src/scene/Post.ts:167`, used as a render target at `theia-panel/src/scene/Post.ts:269` and sampled in the overlay pass at `theia-panel/src/scene/Post.ts:287`. Without resizing, its texture dimensions diverge from `sceneTarget`/`composer` after a viewport change, producing the stretching/cropping symptoms reported.
- DPR handling matches the sibling targets exactly (same `w2 * dpr2, h2 * dpr2` formula as `theia-panel/src/scene/Post.ts:298-300`), so there's no off-by-one or DPR drift introduced.
- `THREE.WebGLRenderTarget.setSize` is a no-op when dimensions are unchanged, so repeated resizes are safe.
- The new line is placed before `composer.setSize` / `bloom.setSize`, consistent with the other targets and the creation order at `theia-panel/src/scene/Post.ts:146-172`.

Edge case worth noting (pre-existing, not introduced here): `renderer.getPixelRatio()` is read at resize time but the renderer's pixel ratio is never explicitly updated on `devicePixelRatio` changes — this is a separate concern from the bug being fixed and applies equally to the existing siblings.

## Code quality
- Style, indentation and ordering match the surrounding lines exactly.
- No comment needed; intent is obvious from context.
- Minor (pre-existing): `w2 * dpr2`, `h2 * dpr2` is repeated four times; could be hoisted into local consts, but that's out of scope for this fix.

## Risk / regressions
Effectively none. The change only enlarges/shrinks an already-allocated render target that was previously stuck at its initial size. It cannot affect render paths that don't go through `ditherTarget`, and the dither/overlay path at `theia-panel/src/scene/Post.ts:269-288` already assumes this target matches scene dimensions.

## Suggestions
None required. Optional nit: extract `const pw = w2 * dpr2, ph = h2 * dpr2;` in a follow-up to deduplicate the four `setSize` calls.

## Verdict
**Approve.**

PR #36, opened 2026-04-25
fix(panel): prevent search bar, filter bar, and side panel overlap on…

### LINKED ISSUE #27: bug(panel): search bar, filter bar, and side panel overlap on narrow viewports

arm64be:
Closes #27.

### DIFF
--- a/theia-panel/src/ui/SearchBar.ts
+++ b/theia-panel/src/ui/SearchBar.ts
@@ -26,7 +26,7 @@ export function createSearchBar(
 
   function applyWrapperStyle() {
     wrapper.style.cssText = `
-      position: absolute; top: 12px; left: 12px;
+      position: absolute; top: 12px; right: 12px;
       z-index: 10; font: 13px/1.4 'Mondwest', var(--theia-font, ui-monospace, monospace);
       color: #${theme.fg}; max-width: 320px; width: calc(100% - 24px);
     `;
--- a/theia-panel/src/ui/SearchBar.ts
+++ b/theia-panel/src/ui/SearchBar.ts
@@ -26,9 +26,9 @@ export function createSearchBar(
 
   function applyWrapperStyle() {
     wrapper.style.cssText = `
-      position: absolute; top: 12px; left: 50%; transform: translateX(-50%);
+      position: absolute; top: 12px; left: 12px;
       z-index: 10; font: 13px/1.4 'Mondwest', var(--theia-font, ui-monospace, monospace);
-      color: #${theme.fg}; width: min(320px, 50vw);
+      color: #${theme.fg}; max-width: 320px; width: calc(100% - 24px);
     `;
   }
 
--- a/theia-panel/src/ui/SidePanel.ts
+++ b/theia-panel/src/ui/SidePanel.ts
@@ -53,8 +53,10 @@ export function createSidePanel(
   let currentId: string | null = null;
 
   function applyPanelStyle() {
+    const narrow = window.matchMedia("(max-width: 900px)").matches;
+    const panelWidth = narrow ? "min(320px, 85vw)" : "min(420px, 45vw)";
     el.style.cssText = `
-      position: absolute; top: 0; right: 0; bottom: 0; width: min(320px, 85vw);
+      position: absolute; top: 0; right: 0; bottom: 0; width: ${panelWidth};
       background: ${themeBgAlpha(theme, 0.95)}; border-left: 1px solid #${theme.border};
       color: #${theme.fg}; font: 13px/1.6 var(--theia-font, ui-monospace, monospace);
       transform: translateX(${currentId ? "0" : "100%"}); transition: transform 220ms ease-out;
--- a/theia-panel/src/ui/SidePanel.ts
+++ b/theia-panel/src/ui/SidePanel.ts
@@ -54,7 +54,7 @@ export function createSidePanel(
 
   function applyPanelStyle() {
     el.style.cssText = `
-      position: absolute; top: 0; right: 0; bottom: 0; width: min(420px, 45vw);
+      position: absolute; top: 0; right: 0; bottom: 0; width: min(320px, 85vw);
       background: ${themeBgAlpha(theme, 0.95)}; border-left: 1px solid #${theme.border};
       color: #${theme.fg}; font: 13px/1.6 var(--theia-font, ui-monospace, monospace);
       transform: translateX(${currentId ? "0" : "100%"}); transition: transform 220ms ease-out;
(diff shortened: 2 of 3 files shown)


zebster-cmd (commented):
# Code Review: PR #36 — `fix(panel): prevent search bar, filter bar, and side panel overlap`

## Summary
A 3-file, ±4-line patch that (1) adds `flex-wrap: wrap` to the filter bar, (2) moves the search bar from centered to `left: 12px` with `width: calc(100% - 24px); max-width: 320px`, and (3) shrinks the side panel from `min(420px, 45vw)` to `min(320px, 85vw)`. Compared with PR #30 (126/-20), which extracts a shared `layoutStyles.ts`, sets explicit z-index ordering, anchors the search bar top-*right*, and uses a `@media (max-width: 900px)` breakpoint, this PR is much smaller but trades one overlap for another.

## Correctness
- **Regression at every width**: the filter bar is still pinned at `top: 12px; left: 12px` (`FilterBar.ts:69`). The search bar is now also at `top: 12px; left: 12px` with `z-index: 10` (`SearchBar.ts:28-32` after patch). Since the filter bar has no z-index, the input now sits on top of all four toggles + the model dropdown at every viewport ≤ ~344 px wide, and partially overlaps them up to ~600 px (filter bar ≈ 250–350 px). PR #30 avoids this by anchoring the search bar to `right: 12px` and giving the filter bar `z-index: 11` over the search bar's `10`.
- **Side panel regression on wide screens**: `min(320px, 85vw)` at `SidePanel.ts:57` caps the panel at 320 px on a 1920 px monitor (was 420 px). Issue #27's suggestion (`min(320px, 85vw)`) was scoped to a `≤900px` media query, not unconditional. Long titles/previews truncated by `SUMMARY_MAX_CHARS` will now be reached sooner on desktop.
- **Side panel still overlaps search bar**: search bar `width: calc(100% - 24px)` extends under the panel; nothing reflows when `show()` runs (`SidePanel.ts:90`). PR #30 toggles `.tp-panel-open` and slides the search bar via `right: calc(min(420px,45vw) + 12px)`.
- **Search dropdown still hidden under side panel** for the same reason — `z-index: 10` on the wrapper without an explicit lower z-index on the panel means stacking depends on insertion order, which is brittle.
- **Positive**: `flex-wrap: wrap` (`FilterBar.ts:70`) is correct and matches the issue's suggestion #4.

## CSS / Layout Quality
- `width: calc(100% - 24px); max-width: 320px` (`SearchBar.ts:31`) is fine but `100%` is `100%` of the absolute-positioned containing block (the canvas container), so on a 320 px viewport the input *is* the whole top row.
- Magic numbers (`12`, `24`, `320`, `85vw`, `45vw`) are duplicated across three files — issue #27 already flagged the cross-file coupling. PR #30's central `layoutStyles.ts` is the better pattern.
- No z-index hygiene change; the existing `z-index: 10` on the search bar is now actively harmful (covers filter controls).
- No breakpoint introduced; this PR is "responsive" only via clamps, so behavior between 600 px and 900 px is essentially unchanged from the bug.

## Accessibility
- Mouse/touch users on narrow viewports cannot click any filter toggle (`FilterBar.ts:114-123`) because the search input intercepts hits in the same rect.
- Keyboard tab order follows DOM order (filter → search → panel), which is preserved, but visually focused filter toggles will be hidden under the input — a WCAG 2.4.11 (focus not obscured) concern.
- Screen-reader semantics unchanged; `role="switch"` on toggles still announces correctly (`FilterBar.ts:12-13`).
- Removing the search bar on narrow viewports (PR #30's `display: none`) has its own a11y cost but at least surfaces filter controls.

## Suggestions
1. Anchor the search bar to `right: 12px` (or stack it below a wrapped filter bar on narrow widths) — keep filter bar at `left: 12px` to avoid the overlap this PR introduces.
2. Make the side panel width responsive only below a breakpoint, not globally; keep `min(420px, 45vw)` for ≥ 900 px.
3. Add an explicit z-index stack: filter bar > search bar > side panel, or move the search bar inside the side panel's stacking context when open.
4. Consider adopting PR #30's `layoutStyles.ts` so the three components don't each carry their own positioning constants.
5. Add `aria-hidden`/focus management when the panel is open over the search input.

## Verdict
**Request changes.** The `flex-wrap` fix is good, but moving the search bar to `left: 12px` directly on top of the filter bar is a clear regression that re-creates the same overlap class issue #27 describes (just rotated 90°). The unconditional side-panel shrink also degrades wide-screen UX. PR #30's scoped breakpoint + shared stylesheet approach is closer to correct; either rebase #36 onto that structure or expand it to (a) anchor the search bar away from the filter bar, (b) gate the panel-width change behind a media query, and (c) establish explicit z-index ordering.

theia-panel/src/ui/SearchBar.ts
```diff
@@ -26,9 +26,9 @@ export function createSearchBar(
 
   function applyWrapperStyle() {
     wrapper.style.cssText = `
-      position: absolute; top: 12px; left: 50%; transform: translateX(-50%);
+      position: absolute; top: 12px; left: 12px;
```
zebster-cmd: `top: 12px; left: 12px;` puts the search bar directly on top of the filter bar (which already lives at `top: 12px; left: 12px;` per `FilterBar.ts:applyBarStyle`). At narrow widths the two will collide. Anchoring the search bar to the right keeps it out of the filter bar's way:

```suggestion
      position: absolute; top: 12px; right: 12px;
```

theia-panel/src/ui/SidePanel.ts
```diff
@@ -54,7 +54,7 @@ export function createSidePanel(
 
   function applyPanelStyle() {
     el.style.cssText = `
-      position: absolute; top: 0; right: 0; bottom: 0; width: min(420px, 45vw);
+      position: absolute; top: 0; right: 0; bottom: 0; width: min(320px, 85vw);
```
zebster-cmd: Unconditionally shrinking from `min(420px, 45vw)` → `min(320px, 85vw)` regresses the wide-screen experience: on a 1440-px display the panel collapses from 420 px → 320 px for no good reason, and on tablets it now eats 85 % of the viewport. Prefer keeping the desktop width and only switching to the narrow rule below a breakpoint, e.g.:

```ts
const panelWidth = window.matchMedia('(max-width: 900px)').matches
  ? 'min(320px, 85vw)'
  : 'min(420px, 45vw)';
```

or inject a `@media (max-width: 900px) { .sv-side-panel { width: min(320px, 85vw); } }` rule alongside the inline style.

zebster-cmd (commented):
Two inline layout notes.

PR #37, opened 2026-04-25
feat(panel): add focus mode to isolate connected nodes on selection —…

### LINKED ISSUE #26: feat(panel): focus mode — isolate connected nodes on selection
## Feature Request

When a node is selected, offer an action that temporarily hides all non-connected nodes and edges so the user can study the local neighborhood without visual clutter.

### Motivation

The current "View in Sessions" button placement only makes sense if the primary action after selecting a node is to isolate its connected subgraph. Right now the graph can be dense, making it hard to trace which edges belong to a specific session.

### Proposed behavior

- Add a "Focus mode" toggle inside the panel's side panel (next to the node metadata).
- When enabled: fade out or hide all nodes/edges that are not directly connected to the selected node.
- When disabled: restore the full graph view.
- The toggle should be stateful per selection (i.e. if I select another node while focus mode is on, it updates to show that node's neighborhood).

### UI suggestion

- A small checkbox or eye-icon toggle labeled "Focus on connected nodes" inside the side panel.
- Or a keyboard shortcut (e.g. `F`) to toggle focus mode.

### Environment

arm64be:
Closes #26.

### DIFF
--- a/theia-panel/src/ui/SidePanel.ts
+++ b/theia-panel/src/ui/SidePanel.ts
@@ -187,6 +187,10 @@ export function createSidePanel(
   function hide() {
     currentId = null;
     el.style.transform = "translateX(100%)";
+    if (focusMode) {
+      focusMode = false;
+      onFocusToggle?.(false);
+    }
   }
 
   function currentNodeId() {
--- a/theia-panel/src/ui/SidePanel.ts
+++ b/theia-panel/src/ui/SidePanel.ts
@@ -47,10 +47,12 @@ export function createSidePanel(
   container: HTMLElement,
   initialTheme: ThemeTokens,
   onNavigate?: (nodeId: string) => void,
+  onFocusToggle?: (enabled: boolean) => void,
 ) {
   let theme = initialTheme;
   const el = document.createElement("aside");
   let currentId: string | null = null;
+  let focusMode = false;
 
   function applyPanelStyle() {
     el.style.cssText = `
@@ -128,11 +130,22 @@ export function createSidePanel(
         <dt style="${dtStyle}">Messages</dt><dd style="margin:0">${node.message_count ?? "-"}</dd>
       </dl>
       <div style="margin:20px 0 0;border-top:1px solid #${theme.border}"></div>
-      <h4 style="margin:10px 0 8px;font-size:10px;letter-spacing:0.12em;opacity:0.5;text-transform:uppercase">Connections</h4>
+      <label style="display:flex;align-items:center;gap:6px;padding:8px 0;font-size:11px;cursor:pointer;user-select:none;color:#${theme.fg}">
+        <input type="checkbox" data-focus-toggle ${focusMode ? 'checked' : ''} style="accent-color:#${theme.accent}">
+        Focus on connected nodes
+      </label>
+      <h4 style="margin:0 0 8px;font-size:10px;letter-spacing:0.12em;opacity:0.5;text-transform:uppercase">Connections</h4>
       <div style="display:flex;flex-direction:column;gap:8px">
         ${relatedEdges.length === 0 ? '<div style="opacity:0.5;font-size:11px">No connections</div>' : relatedEdges.map((e) => renderEdge(node, e, theme, !!onNavigate)).join("")}
       </div>
     `;
+    const focusToggle = el.querySelector("[data-focus-toggle]") as HTMLInputElement;
+    if (focusToggle) {
+      focusToggle.onchange = () => {
+        focusMode = focusToggle.checked;
+        onFocusToggle?.(focusMode);
+      };
+    }
     (el.querySelector("#sv-close") as HTMLButtonElement).onclick = hide;
   }
 
--- a/theia-panel/src/ui/SidePanel.ts
+++ b/theia-panel/src/ui/SidePanel.ts
@@ -131,7 +131,7 @@ export function createSidePanel(
       </dl>
       <div style="margin:20px 0 0;border-top:1px solid #${theme.border}"></div>
       <label style="display:flex;align-items:center;gap:6px;padding:8px 0;font-size:11px;cursor:pointer;user-select:none;color:#${theme.fg}">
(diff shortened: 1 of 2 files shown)


zebster-cmd (commented):
# Review — PR #37: Focus mode

## Summary
Adds a "Focus on connected nodes" checkbox to the `SidePanel`. When enabled, selecting a node calls a new `applyFocusModeIfEnabled` that hides every node not directly adjacent to the selection and rebuilds the edge layer with that subset. Plumbing: `SidePanel.createSidePanel` gains an `onFocusToggle` callback; `index.ts` keeps a `focusEnabled` flag and re-invokes the helper from all three selection paths (panel navigate, mouseup, reload).

## Correctness
- Traversal is **1-hop only** — no BFS/DFS. That matches issue #26 ("directly connected") so it's fine, but the PR title says "connected nodes" which usually implies the whole component. Worth clarifying in the PR body.
- Edges are treated undirected (`source === id || target === id`) at `theia-panel/src/index.ts:201-208`, which is consistent with how `related` is computed elsewhere (e.g. `theia-panel/src/index.ts:86-88`).
- Kind filter is honored in the neighbor walk (`theia-panel/src/index.ts:202`), and `edges.rebuild` is fed a `filteredNodeIndex` restricted to visible neighbors (`theia-panel/src/index.ts:214-220`) — orphan edges won't render.
- Perf: O(E + N) per click; full edge buffer rebuild on every selection while focus is on. Acceptable for current graph sizes but worth memoizing per `(selectedId, kinds, modelFilter)` if graphs grow.
- **Bug — filter desync:** `updateVisibility()` (`theia-panel/src/index.ts:160-195`) is invoked when kinds/model filters change (`theia-panel/src/index.ts:446-450`). It restores all `visibleNodeIds` but does not re-apply focus mode, so the checkbox stays "on" while the graph is fully unfocused until the next click.

## UX
- Toggle placement matches the issue's UI suggestion.
- **No escape hatch from the close button**: `SidePanel.hide()` (`theia-panel/src/ui/SidePanel.ts:187-190`) does not fire `onFocusToggle(false)` nor restore visibility. Closing the panel via `×` or `Esc` (`theia-panel/src/ui/SidePanel.ts:161-163`) leaves the graph crippled — nodes stay hidden with the panel gone. This is the most user-visible regression.
- Issue #26 also suggested an `F` keyboard shortcut — not implemented.
- The PR uses hard `setVisible(false)` rather than the "fade out" alternative the issue proposed; spatial context is lost. Opacity-based de-emphasis (e.g. `nodes.setHighlight`-like attenuation) would be friendlier and match the existing hover semantics.

## Code quality
- `applyFocusModeIfEnabled` (`theia-panel/src/index.ts:197-221`) duplicates the edge-rebuild tail of `updateVisibility` — extract a shared `rebuildEdgesFor(visibleSet)` helper.
- Two sources of truth for the toggle state: `focusEnabled` in `index.ts:63` and `focusMode` in `SidePanel.ts:55`. Functional today, but easy to drift; passing state in via `show()` or returning a getter would be cleaner.
- Inline `onchange` re-bound on every `renderContent` (`theia-panel/src/ui/SidePanel.ts:142-148`) — fine, but worth a `data-` delegated handler for consistency with the existing nav delegation.

## Risk vs PR #41
PR #41 also extends `createSidePanel`'s signature with a 4th-positional callback (`onClose`) and introduces persistent `selectedIdx` highlight. **Direct conflict**: both PRs add a different 4th parameter and both wrap the panel construction at `theia-panel/src/index.ts:76-102`. Whichever lands second will need a non-trivial rebase; recommend converting to an options bag (`{ onNavigate, onClose, onFocusToggle }`) before merge. Behaviorally the two are compatible — focus hides nodes, #41 colors the selected one, both via instanced attributes.

## Suggestions
1. In `SidePanel.hide()`, call `onFocusToggle?.(false)` (or the merged `onClose`) so closing restores the graph.
2. In the filter bar callback (`theia-panel/src/index.ts:446-450`), call `applyFocusModeIfEnabled(sidePanel.currentNodeId())` after `updateVisibility()`.
3. Add `F` keyboard shortcut on the panel/container.
4. Consider opacity fade instead of `setVisible(false)` for de-emphasis.
5. Refactor to options-bag callback signature to unblock #41.

## Verdict
**Request changes.** Logic is sound and minimal, but the close-button escape-hatch gap and the filter-toggle desync are real regressions, and the SidePanel signature collides with PR #41. Address (1) and (2) above and coordinate the signature with #41 before merge.

theia-panel/src/ui/SidePanel.ts
```diff
@@ -47,10 +47,12 @@ export function createSidePanel(
   container: HTMLElement,
   initialTheme: ThemeTokens,
   onNavigate?: (nodeId: string) => void,
+  onFocusToggle?: (enabled: boolean) => void,
```
zebster-cmd, line 50: Heads-up: PR #41 (selected-node highlighting) adds an `onClose?: () => void` parameter at this exact 4th-arg position. Whichever lands second will silently bind its callback to the wrong slot. Strongly recommend converting `createSidePanel` to take an options bag before either merges:

```ts
export function createSidePanel(
  container: HTMLElement,
  initialTheme: ThemeTokens,
  callbacks: {
    onNavigate?: (nodeId: string) => void;
    onFocusToggle?: (enabled: boolean) => void;
    onClose?: () => void;
  } = {},
) { ... }
```

Also: while `hide()` isn't shown in this hunk, please make sure it calls `onFocusToggle?.(false)` so closing the panel restores graph visibility — otherwise the `focusEnabled` flag in `index.ts` stays true and unrelated nodes remain hidden after the user dismisses the panel. And the kinds/model filter callback in `index.ts` should call `applyFocusModeIfEnabled(sidePanel.currentNodeId())` after `updateVisibility()` to avoid filter desync.

zebster-cmd (commented):
Inline notes on the focus-mode wiring.

zebster-cmd:
Follow-up bug filed: #58 — clicking the focus-mode checkbox inside the side panel triggers the canvas mouseup handler and closes the panel. Suggested fix: either add an event-bubbling guard in index.ts or move the toggle to the filter bar.

PR #38, opened 2026-04-25
ui(plugin): collapse selected session card into compact inline row — …

### LINKED ISSUE #25: ui(plugin): selected session card is visually heavy and poorly positioned
## Bug Report

The selected-node card below the constellation iframe is visually heavy and poorly positioned.

### Current behavior

After clicking a node, a full `Card` with `CardHeader` ("Selected Session" title) + `CardContent` (node ID + button) appears **below** the iframe. This creates:

- Excessive vertical space for a single action.
- The node ID is already visible in the side panel; repeating it here is redundant.
- The "View in Sessions" button is far from the visual context where the user just clicked.

### Suggested improvements

1. **Collapse into a compact inline row** (badge + small button) instead of a full card.
2. **Move the action into the iframe's own side panel** so the button lives next to the node metadata.
3. **Or place it as a floating action bar** directly under the constellation viewport.

### Environment

arm64be:
Closes #25.

### DIFF
--- a/plugin/src/index.js
+++ b/plugin/src/index.js
@@ -305,8 +305,13 @@
       ),
 
       // Selected node — compact inline row
-      selectedNode && h("div", { className: "flex items-center gap-2" },
-        h(Badge, { variant: "outline", className: "font-courier text-xs" }, selectedNode),
+      selectedNode && h("div", {
+        role: "status",
+        "aria-label": "Selected session " + selectedNode,
+        "data-testid": "selected-session-row",
+        className: "flex items-center gap-2",
+      },
+        h(Badge, { variant: "outline", className: "font-courier text-xs truncate max-w-[24ch]" }, selectedNode),
         h(Button, {
           onClick: function () {
             window.location.hash = "#/sessions?id=" + selectedNode;
--- a/plugin/src/index.js
+++ b/plugin/src/index.js
@@ -304,22 +304,15 @@
         })
       ),
 
-      // Selected node info
-      selectedNode && h(Card, null,
-        h(CardHeader, null,
-          h(CardTitle, { className: "text-base" }, "Selected Session")
-        ),
-        h(CardContent, null,
-          h("div", { className: "flex items-center gap-3" },
-            h("span", { className: "font-courier text-sm" }, selectedNode),
-            h(Button, {
-              onClick: function () {
-                window.location.hash = "#/sessions?id=" + selectedNode;
-              },
-              variant: "outline", size: "sm",
-            }, "View in Sessions")
-          )
-        )
+      // Selected node — compact inline row
+      selectedNode && h("div", { className: "flex items-center gap-2" },
+        h(Badge, { variant: "outline", className: "font-courier text-xs" }, selectedNode),
+        h(Button, {
+          onClick: function () {
+            window.location.hash = "#/sessions?id=" + selectedNode;
+          },
+          variant: "outline", size: "sm",
+        }, "View in Sessions")
       )
     );
   }


zebster-cmd (commented):
# Code Review — PR #38: collapse selected session card into compact inline row

## Summary
Minimal, well-scoped change (9/-16) on a single file. Replaces the heavy `Card` + `CardHeader("Selected Session")` + `CardContent` block at `plugin/src/index.js:307-323` with an inline flex row containing an outline `Badge` (the session id) and the "View in Sessions" `Button`. Directly implements suggestion #1 of the linked issue ("Collapse into a compact inline row (badge + small button)").

## Correctness
- All behavioural surface is preserved: `selectedNode` is still rendered and the click target still navigates via `window.location.hash = "#/sessions?id=" + selectedNode` (`plugin/src/index.js:316-318` post-patch).
- The only piece dropped is the redundant "Selected Session" header text — issue #25 explicitly notes this is duplicated by the side panel, so the loss is intentional.
- **Truncation:** session IDs are arbitrary strings rendered inside a `Badge` with `font-courier text-xs` and no `truncate` / `min-w-0` / `max-w-…` classes. In a narrow viewport (or fullscreen on small windows) a long UUID-style id can push the button off-row or force overflow. The previous version had the same risk (`font-courier text-sm` span), so this is not a regression — but it is also not fixed.

## UX / Accessibility
- Visual goal (compact, less vertical weight, button closer to context) is achieved.
- The implicit "Selected Session" label is removed and **not replaced by an `aria-label` / `aria-live` region / heading**. Screen-reader users lose the only textual cue identifying what the badge represents. Suggest `h("div", { role: "status", "aria-label": "Selected session " + selectedNode, className: "flex items-center gap-2" }, …)` or wrap with a visually-hidden label.
- `Button` still uses `size: "sm"` — already pre-existing, but combined with the now-tighter row this may dip below comfortable touch-target size on mobile. Worth confirming with design.
- Keyboard navigation is unchanged — the `Button` remains the only focusable element; tab order is fine.

## Code Quality
- Idiomatic with the surrounding hyperscript style. `Badge`, `Button` already imported (`plugin/src/index.js:29`, used at `:277-279`).
- `gap-2` (was `gap-3`) is consistent with the header action group at `plugin/src/index.js:283`.
- No dead imports introduced; `CardHeader`/`CardTitle`/`CardContent` still used by the page header (`:272-294`).

## Risk — interaction with PR #40
**Direct conflict.** PR #40 rewrites the same `onClick` body (`plugin/src/index.js:317`) to `window.location.href = "/sessions?resume=" + selectedNode`. Both PRs touch overlapping lines, so whichever merges second will hit a textual conflict. The conflict is trivial to resolve (apply #40's new href inside #38's compact row), but reviewers should sequence the merges and re-target #40 onto this branch's button.

## Suggestions
1. Add `truncate max-w-[24ch]` (or `min-w-0` on a wrapping span) to the badge to defend against long ids.
2. Add `role="status"` + `aria-label` so the row is announced as the selected session.
3. Consider `data-testid="selected-session-row"` to ease future tests.

## Verdict
**Approve with minor comments.** The change is small, faithful to issue #25, and behaviour-preserving. Truncation hardening and an `aria-label` would make it a clean ship; merge ordering vs PR #40 must be coordinated.

zebster-cmd (commented):
Two small a11y / robustness suggestions on the compact row.

plugin/src/index.js
```diff
@@ -304,22 +304,15 @@
...
-            h(Button, {
-              onClick: function () {
-                window.location.hash = "#/sessions?id=" + selectedNode;
-              },
-              variant: "outline", size: "sm",
-            }, "View in Sessions")
-          )
-        )
+      // Selected node — compact inline row
+      selectedNode && h("div", { className: "flex items-center gap-2" },
```
zebster-cmd: When the row appears mid-DOM after a graph click, screen readers don't announce it. Adding `role="status"` plus an `aria-label` makes the change observable, and gives the badge its own accessible name when collapsed:

```suggestion
      selectedNode && h("div", { className: "flex items-center gap-2", role: "status", "aria-label": `Selected session ${selectedNode}` },
```

plugin/src/index.js
```diff
@@ -304,22 +304,15 @@
...
+        h(Badge, { variant: "outline", className: "font-courier text-xs" }, selectedNode),
```
zebster-cmd: Hermes session IDs can be quite long (UUIDs, hashes, etc.). Without a width cap this row will push the *View in Sessions* button off-screen on narrow dashboards. Truncate the badge with a sane max width:

```suggestion
        h(Badge, { variant: "outline", className: "font-courier text-xs truncate max-w-[24ch] min-w-0" }, selectedNode),
```

PR #40, opened 2026-04-25
fix(plugin): fix 'View in Sessions' navigation to use correct dashboa…

### LINKED ISSUE #24: bug(plugin): 'View in Sessions' button does not navigate
## Bug Report

The 'View in Sessions' button in the selected-node card below the constellation iframe is non-functional.

### Root cause

The button's  handler sets:

This fails for two reasons:

1. **Routing mismatch** — The Hermes dashboard uses  (react-router-dom), not hash-based routing. Setting  does not trigger a route change; it only appends  to the current URL while the page stays on .

2. **Missing query param handler** — Even if navigation worked,  does not read an uid=1000(zebkuyper) gid=1000(zebkuyper) groups=1000(zebkuyper),4(adm),20(dialout),24(cdrom),25(floppy),27(sudo),29(audio),30(dip),44(video),46(plugdev),116(netdev),999(docker) query parameter. The sessions page only supports  for the chat resumption flow.

### Repro

2. Click any node to open the side panel.
3. Scroll down to the 'Selected Session' card below the iframe.
4. Click 'View in Sessions'.
5. **Expected:** Navigate to the session detail/list view.
6. **Actual:** Nothing happens. URL hash changes but page stays on .

### Suggested fixes

- Navigate to  (matches existing dashboard behavior for resuming a session).
- Or expose a navigation helper on  so plugins can call into the router properly.

### Environment

- Branch:  (commit )
- Hermes dashboard with BrowserRouter

arm64be:
Closes #24.

### DIFF
--- a/plugin/src/index.js
+++ b/plugin/src/index.js
@@ -314,7 +314,7 @@
             h("span", { className: "font-courier text-sm" }, selectedNode),
             h(Button, {
               onClick: function () {
-                window.location.hash = "#/sessions?id=" + selectedNode;
+                window.location.href = "/sessions?resume=" + selectedNode;
               },
               variant: "outline", size: "sm",
             }, "View in Sessions")
--- a/plugin/src/index.js
+++ b/plugin/src/index.js
@@ -314,7 +314,10 @@
             h("span", { className: "font-courier text-sm" }, selectedNode),
             h(Button, {
               onClick: function () {
-                window.location.href = "/sessions?resume=" + selectedNode;
+                // Navigate to sessions page with resume session id (see issue #24)
+                var url = "/sessions?resume=" + encodeURIComponent(selectedNode);
+                history.pushState({}, "", url);
+                window.dispatchEvent(new PopStateEvent("popstate"));
               },
               variant: "outline", size: "sm",
             }, "View in Sessions")


zebster-cmd (commented):
# PR #40 Review — `fix(plugin): View in Sessions navigation`

## Summary
Single-line change in `plugin/src/index.js:317` swaps the broken hash-based navigation for a real URL navigation matching the dashboard's existing resume route, per issue #24's suggested fix.

```diff
- window.location.hash = "#/sessions?id=" + selectedNode;
+ window.location.href = "/sessions?resume=" + selectedNode;
```

## Correctness
- **Route**: `/sessions?resume=<id>` matches the contract the dashboard already supports for chat resumption (per issue #24 body). The previous `#/sessions?id=` form never triggered a route change under BrowserRouter — so this fixes the reported defect.
- **Session id**: `selectedNode` is the same value rendered in the card label (`plugin/src/index.js:314`), so the id passed downstream is consistent with what the user clicked.
- **Deep-linking**: query param key changes (`id` → `resume`). That's correct because only `resume` is wired up in the sessions page; no other consumer of the old `id` param exists in this repo.

## Code quality
- `selectedNode` is interpolated raw — should be `encodeURIComponent(selectedNode)` (`plugin/src/index.js:317`). Session ids in this codebase look safe today, but a single `&`/`#`/space breaks the URL silently.
- `window.location.href = "/sessions?..."` triggers a **full page reload**, losing BrowserRouter state and re-mounting the iframe. Issue #24 itself suggests an alternative ("expose a navigation helper on `window` so plugins can call into the router"); a `history.pushState` + `popstate` dispatch (or an injected `navigate()` helper) would be smoother. Acceptable as a minimal fix, worth a follow-up.
- Hard-coded absolute path `/sessions` assumes the dashboard is mounted at root; would break under a non-root base path. Not a regression vs. the previous code.

## Risk
- **Conflicts with PR #38** (`plugin/src/index.js:307-322`): #38 refactors this same block into a compact inline row but **retains the buggy `window.location.hash = "#/sessions?id="`** line. Whichever merges second will need a rebase, and #38 must adopt the #40 navigation or it will silently regress this bug. Recommend coordinating merge order or asking #38's author to cherry-pick the new `href`/`resume` form.
- No new tests; manual verification per issue #24 repro steps is required.

## Suggestions
1. `encodeURIComponent(selectedNode)`.
2. Consider `history.pushState({}, "", url); window.dispatchEvent(new PopStateEvent("popstate"))` to avoid full reload.
3. Add a brief comment at `plugin/src/index.js:316` referencing issue #24 + the `resume` contract.
4. Coordinate with PR #38 to keep the fix.

## Verdict
**Approve with minor suggestions.** Correctly addresses #24; URL-encoding and the #38 collision should be handled before/at merge.

plugin/src/index.js
```diff
@@ -314,7 +314,7 @@
             h("span", { className: "font-courier text-sm" }, selectedNode),
             h(Button, {
               onClick: function () {
-                window.location.hash = "#/sessions?id=" + selectedNode;
+                window.location.href = "/sessions?resume=" + selectedNode;
```
zebster-cmd: `selectedNode` is the raw node id and may legitimately contain `&`, `#`, `+`, spaces, or other characters that break query-string parsing. URL-encode it:

```suggestion
                window.location.href = "/sessions?resume=" + encodeURIComponent(selectedNode);
```

zebster-cmd: Separately: `window.location.href = ...` triggers a full page reload, which re-mounts the dashboard's iframe, throws away the WebGL context built by `theia-panel`, and is visibly janky. If the dashboard uses a client-side router (and it does — `#/sessions` was the previous form), prefer a soft navigation that the router can pick up:

```js
const url = "/sessions?resume=" + encodeURIComponent(selectedNode);
history.pushState({}, "", url);
window.dispatchEvent(new PopStateEvent("popstate"));
```

That keeps the constellation alive across the navigation.

zebster-cmd (commented):
Two notes on the new hard-redirect.

zebster-cmd:
Regression filed: #60 — the fix from this PR was overwritten by PR #38 and the subsequent restore commit (1a34fa1) introduced a new broken approach instead of restoring the actual  fix.

PR #41, opened 2026-04-25
feat(panel): add persistent selection highlight separate from hover —…

### LINKED ISSUE #21: Selected node has no persistent visual highlight — hover and selection are conflated
## Problem

When a node is clicked and the side panel opens, there is no persistent visual indication of which node is selected in the constellation. Moving the mouse away from the node immediately removes all highlighting, so the user loses track of which session they are inspecting.

Hover and selection are conflated into a single state, which means:
1. Clicking a node opens the side panel but the node does not stay highlighted.
2. Moving the mouse to another node (or off the canvas) clears the highlight entirely.
3. There is no way to visually correlate the open side panel with a specific node in the graph.

## Reproduction

2. Click any node → side panel opens with session details.
3. Move the mouse slightly away from the node.
4. The node returns to its base amber color. There is no visual marker showing which node is "selected."

## Root cause

In `theia-panel/src/scene/Nodes.ts`, there is only one `highlighted` `Set<number>` that serves **both** hover and any would-be selection state:

```typescript
const highlighted = new Set<number>();

setHighlight(i, on) {
  if (on) highlighted.add(i);
  else highlighted.delete(i);
}
```

In `theia-panel/src/index.ts`, the hover callback unconditionally clears every node's highlight except the currently hovered one:

```typescript
picker.onHover((idx) => {
  // ...
  if (idx === null) {
    tooltip.hide();
  } else {
    nodes.setHighlight(idx, true);   // hover highlight
    // ...
  }
  for (let i = 0; i < nodes.count; i++) {
    if (i !== idx) nodes.setHighlight(i, false);  // ← wipes EVERYTHING
  }
  // ...
});
```

The click handler (`mouseup`) opens `sidePanel.show()` but **never** calls `nodes.setHighlight()` to mark the node as selected. Even if it did, the hover callback would wipe it on the next mouse move.

## Expected behavior

- **Hover**: a subtle highlight (current white glow is fine) while the cursor is over a node.
- **Selection**: a distinct, persistent highlight color on the node whose details are currently shown in the side panel. This color should remain even when the mouse moves elsewhere.
- The side panel's close action should clear the selection highlight.

### 1. Add a `selected` state separate from `highlighted` in `NodeLayer`

```typescript
// Nodes.ts
const highlighted = new Set<number>();
const selected = new Set<number>();

// ...
setHighlight(i, on) { /* same */ }
setSelected(i, on) {
  if (on) selected.add(i);
  else selected.delete(i);
}
```

### 2. Add a `nodeSelected` color to the palette

```typescript
// aesthetic.ts
export const PALETTE = {
  // ...
  nodeHighlight: 0xffffff,    // hover (keep current)
  nodeSelected: 0x66d9ef,     // selected — cyan to contrast with amber base
  // ...
};
```

### 3. Render selected nodes with priority in `setTime()`

```typescript
setTime(t: number) {
  const colorAttr = mesh.instanceColor!;
  for (let i = 0; i < n; i++) {
    if (selected.has(i)) {
      colorAttr.setXYZ(i, selectedColor.r, selectedColor.g, selectedColor.b);
      continue;
    }
    if (highlighted.has(i)) {
      colorAttr.setXYZ(i, highlightColor.r, highlightColor.g, highlightColor.b);
      continue;
    }
    // ... base color + twinkle
  }
  colorAttr.needsUpdate = true;
}
```

### 4. Wire selection in `index.ts`

- In the `mouseup` click handler: `nodes.setSelected(idx, true)` and `nodes.setHighlight(idx, false)` (to avoid double-color conflict).
- In `picker.onHover`: only mutate `highlighted`, never touch `selected`.
- In `sidePanel.hide()` (or close button): `nodes.setSelected(idx, false)`.

## Environment

- Hermes Agent: latest main
- Theia: main branch
- Browser: Chrome / Firefox
- OS: WSL2

arm64be:
Closes #21.

### DIFF
--- a/theia-panel/src/index.ts
+++ b/theia-panel/src/index.ts
@@ -76,29 +76,37 @@ export async function mount(
   function clearSelected() {
     if (selectedIdx !== null) {
       nodes.setSelected(selectedIdx, false);
+      nodes.setHighlight(selectedIdx, false);
       selectedIdx = null;
     }
   }
 
-  const sidePanel = createSidePanel(element, theme, (targetId) => {
-    const idx = nodeIndex.get(targetId);
-    if (idx === undefined || !visibleNodeIds.has(targetId)) return;
-    const sn = simNodes[idx];
-    if (!sn) return;
+  function select(idx: number) {
     clearSelected();
     selectedIdx = idx;
     nodes.setSelected(idx, true);
     nodes.setHighlight(idx, false);
-    ctx.focusOn(sn.x, sn.y, 1.5);
-    const n = currentGraph.nodes[idx]!;
-    const related = currentGraph.edges.filter(
-      (e) => (e.source === n.id || e.target === n.id) && kinds.has(e.kind),
-    );
-    sidePanel.show(n, related);
-    emit("node-click", targetId);
-  }, () => {
-    clearSelected();
-    nodes.flush();
+  }
+
+  const sidePanel = createSidePanel(element, theme, {
+    onNavigate: (targetId) => {
+      const idx = nodeIndex.get(targetId);
+      if (idx === undefined || !visibleNodeIds.has(targetId)) return;
+      const sn = simNodes[idx];
+      if (!sn) return;
+      select(idx);
+      ctx.focusOn(sn.x, sn.y, 1.5);
+      const n = currentGraph.nodes[idx]!;
+      const related = currentGraph.edges.filter(
+        (e) => (e.source === n.id || e.target === n.id) && kinds.has(e.kind),
+      );
+      sidePanel.show(n, related);
+      emit("node-click", targetId);
+    },
+    onClose: () => {
+      clearSelected();
+      nodes.flush();
+    },
   });
 
   let lastMouse = { x: 0, y: 0 };
@@ -384,16 +392,16 @@ export async function mount(
     if (isMouseDown && !hasDragged && performance.now() - lastWheelAt >= 200) {
       const idx = picker.pickAt(e.clientX, e.clientY, 1.0);
       if (idx !== null) {
-        clearSelected();
-        selectedIdx = idx;
-        nodes.setSelected(idx, true);
-        nodes.setHighlight(idx, false);
+        select(idx);
         const n = currentGraph.nodes[idx]!;
         const related = currentGraph.edges.filter(
           (e) => (e.source === n.id || e.target === n.id) && kinds.has(e.kind),
         );
         sidePanel.show(n, related);
         emit("node-click", n.id);
+      } else {
+        clearSelected();
+        sidePanel.hide();
       }
     }
     isMouseDown = false;
--- a/theia-panel/src/index.ts
+++ b/theia-panel/src/index.ts
@@ -70,13 +70,25 @@ export async function mount(
   let simulation: ReturnType<typeof createSimulation>["simulation"];
   let picker: ReturnType<typeof createPicker>;
   let searchBar: ReturnType<typeof createSearchBar>;
+  let selectedIdx: number | null = null;
 
   const tooltip = createTooltip(element, theme);
+  function clearSelected() {
+    if (selectedIdx !== null) {
+      nodes.setSelected(selectedIdx, false);
+      selectedIdx = null;
+    }
+  }
+
   const sidePanel = createSidePanel(element, theme, (targetId) => {
     const idx = nodeIndex.get(targetId);
     if (idx === undefined || !visibleNodeIds.has(targetId)) return;
     const sn = simNodes[idx];
     if (!sn) return;
+    clearSelected();
+    selectedIdx = idx;
+    nodes.setSelected(idx, true);
+    nodes.setHighlight(idx, false);
     ctx.focusOn(sn.x, sn.y, 1.5);
     const n = currentGraph.nodes[idx]!;
     const related = currentGraph.edges.filter(
@@ -84,6 +96,9 @@ export async function mount(
     );
     sidePanel.show(n, related);
     emit("node-click", targetId);
+  }, () => {
+    clearSelected();
+    nodes.flush();
   });
 
   let lastMouse = { x: 0, y: 0 };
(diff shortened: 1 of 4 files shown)


zebster-cmd (commented):
# Review — PR #41: Persistent selection highlight

## Summary
Adds independent `selected` state on the node layer (`Nodes.ts:124`, `Nodes.ts:158`, `Nodes.ts:179-182`, `Nodes.ts:199-207`) coloured by `PALETTE.nodeSelected` (`aesthetic.ts:6`) and a `selectedIdx` tracker in `index.ts` updated on side-panel navigate, click, and reload, with a new `onClose` callback from `SidePanel.hide()` (`SidePanel.ts:178`) used to clear it. Hover loop is reworked so the selected node is exempt from highlight-clearing (`index.ts:282-287` in diff).

## Correctness
- Selection persists across mouse-leave because the hover loop now skips `selectedIdx` (`index.ts` diff @@ -231,15 +246,17). Good.
- **Stale highlight bug**: when the user hovers the selected node, `setHighlight(selectedIdx,true)` runs; on mouse-leave the new guard `i !== selectedIdx` prevents clearing, leaving `highlighted` set true. After `clearSelected()` (panel close), the node renders in highlight white instead of base — `Nodes.ts:setTime` falls through to the `highlighted` branch. Fix: have `clearSelected()` also call `setHighlight(selectedIdx,false)`.
- Background click does **not** deselect (`index.ts:407-423`); only the close button does. Issue #21 mentions clicking off the node — worth confirming intended UX.
- No multi-select; not requested.
- `reload()` calls `clearSelected()` after `setupGraph` rebuilds the node layer (`index.ts:506-509`); harmless because the new `selected` set is empty, but `selectedIdx` is stale until reset — fragile.
- Three near-identical `clearSelected(); selectedIdx=…; setSelected(true); setHighlight(false)` blocks (sidePanel cb, mouseup, reload) — extract a `select(idx)` helper.

## Visual / a11y
`0x66d9ef` is the same hue as `edgeSearch` (`aesthetic.ts:8`), so a selected node visually rhymes with cross-search edges — possible confusion. Contrast vs amber base and white hover is fine; cyan on `0x07080d` is comfortably AA. Consider a ring/outline rather than a hue swap to avoid colliding with edge semantics.

## Risk vs PR #37 (focus mode)
Both PRs add a 4th positional arg to `createSidePanel` (`SidePanel.ts:50`): #37 → `onFocusToggle`, #41 → `onClose`. **Merge conflict guaranteed**; whichever lands second must switch to an options object or a 5th param. Behavioural overlap: in #37, `hide()` does not disable focus mode or call `updateVisibility()`; combined with #41's `onClose → clearSelected`, closing the panel while focus is on leaves the graph filtered to the (now unselected) node's neighbours. Either `onClose` should also reset focus mode or focus should track `selectedIdx`.

## Suggestions
1. Clear `highlighted` for old `selectedIdx` inside `clearSelected()`.
2. Extract `select(idx)` helper.
3. Deselect on empty-canvas click.
4. Use options bag for `createSidePanel` callbacks to avoid #37 conflict.
5. Consider an outline/scale cue instead of reusing the cyan edge hue.

## Verdict
**Request changes** — feature works directionally and matches issue #21's intent, but the highlight-leak bug and the SidePanel signature collision with PR #37 should be resolved before merge.

theia-panel/src/index.ts
```diff
@@ -70,20 +70,35 @@ export async function mount(
   let simulation: ReturnType<typeof createSimulation>["simulation"];
   let picker: ReturnType<typeof createPicker>;
   let searchBar: ReturnType<typeof createSearchBar>;
+  let selectedIdx: number | null = null;
 
   const tooltip = createTooltip(element, theme);
+  function clearSelected() {
+    if (selectedIdx !== null) {
+      nodes.setSelected(selectedIdx, false);
```
zebster-cmd, line 78: When a node is hovered *and* selected, it lives in both the `highlighted` and `selected` sets simultaneously (the hover handler skips clearing highlight on `selectedIdx` — see the `else if (i !== selectedIdx)` branch below). After `clearSelected()` runs, the previous selection still has its highlight flag set, so `setTime` will keep painting it in the highlight colour until the cursor moves. Drop the highlight at the same time:

```suggestion
      nodes.setSelected(selectedIdx, false);
      nodes.setHighlight(selectedIdx, false);
```

theia-panel/src/index.ts
```diff
@@ -367,6 +384,10 @@ export async function mount(
     if (isMouseDown && !hasDragged && performance.now() - lastWheelAt >= 200) {
       const idx = picker.pickAt(e.clientX, e.clientY, 1.0);
       if (idx !== null) {
```
zebster-cmd, line 394: There's no `else` branch here — clicking on empty canvas leaves the previously-selected node selected indefinitely. The issue UX ("deselect on click-off") expects the opposite. Add an `else` that calls `clearSelected(); nodes.flush();` so an empty-space click clears the selection. (You'll also want to call `sidePanel.hide()` if it's currently open, but that depends on whether you treat selection and panel visibility as coupled.)

theia-panel/src/ui/SidePanel.ts
```diff
@@ -47,6 +47,7 @@ export function createSidePanel(
   container: HTMLElement,
   initialTheme: ThemeTokens,
   onNavigate?: (nodeId: string) => void,
+  onClose?: () => void,
```
zebster-cmd: Heads-up: PR #37 (focus mode) adds an `onFocusToggle?: (enabled: boolean) => void` parameter at this exact 4th-arg position. Whichever PR lands second will silently bind its callback to the wrong slot at every existing call site. Strongly recommend converting `createSidePanel` to take an options bag before either merges:

```ts
export function createSidePanel(
  container: HTMLElement,
  initialTheme: ThemeTokens,
  callbacks: {
    onNavigate?: (nodeId: string) => void;
    onClose?: () => void;
    onFocusToggle?: (enabled: boolean) => void;
  } = {},
) { ... }
```

zebster-cmd (commented):
Three notes on the selection-state machine.

PR #42, opened 2026-04-25
fix(core): respect HERMES_HOME for profile-aware data paths — fixes #19

### LINKED ISSUE #19: Theia Constellation ignores Hermes profiles — always renders root profile graph
## Problem

Hermes profiles create fully isolated session stores under `~/.hermes/profiles/<name>/state.db`. The dashboard (`hermes dashboard --profile <name>`) correctly scopes itself to that profile's `HERMES_HOME`, so Sessions / Search / Config all show the right data.

However, Theia Constellation hardcodes its data path to `~/.hermes/` via `THEIA_HOME`, completely ignoring the active `HERMES_HOME`. This means:

- `swarm-orchestrator dashboard` shows swarm-orchestrator sessions in the dashboard, but the Constellation tab shows the **root/default** profile's graph.
- `theia-core` with no args always reads `~/.hermes/state.db` instead of the active profile's database.

## Root cause

`theia-core` and `graph_data.py` both resolve paths like this:

```python
def _theia_home() -> Path:
    return Path(os.environ.get("THEIA_HOME", Path.home() / ".hermes"))
```

They know nothing about `HERMES_HOME`, so when a profile is active (`HERMES_HOME=~/.hermes/profiles/swarm-orchestrator`), Theia still looks at the literal `~/.hermes` directory.

## Reproduction

```bash
# 1. Create a profile with some sessions
hermes profile create test --clone
hermes -p test chat -q "hello world"

# 2. Start dashboard under that profile
hermes -p test dashboard --host 0.0.0.0 --insecure --no-open

# 3. Open Constellation tab → it shows the ROOT profile graph (empty or wrong sessions),
#    while the Sessions tab correctly shows the 'test' profile's 'hello world' session.
```

## Expected behavior

When `HERMES_HOME` points to a profile directory, Theia should automatically use that same directory for its `state.db` and `theia-graph.json`, **unless** `THEIA_HOME` is explicitly overridden.

The lookup priority should be:
1. `$THEIA_HOME` (explicit override)
2. `$HERMES_HOME` (active profile, if set)
3. `~/.hermes` (fallback)

### Option A: Respect `HERMES_HOME` in `theia-core` and `graph_data.py`

Change `_theia_home()` in both places to:

```python
def _theia_home() -> Path:
    return Path(os.environ.get("THEIA_HOME",
               os.environ.get("HERMES_HOME",
               Path.home() / ".hermes")))
```

This is a 2-line change that makes Theia "just work" with profiles without breaking existing setups.

### Option B: Aggregate all profiles (future)

A more ambitious fix would be to have the dashboard plugin discover all profile databases under `~/.hermes/profiles/*/state.db` and merge them into a single constellation view with profile-color-coded nodes. This is probably out of scope for a quick fix.

## Environment

- Hermes Agent: latest main
- Theia: main branch (commit 16e78d3+)
- OS: WSL2 / Ubuntu

---

**Workaround for users hitting this now:**

```bash
THEIA_HOME=~/.hermes/profiles/swarm-orchestrator theia-core
THEIA_HOME=~/.hermes/profiles/swarm-orchestrator swarm-orchestrator dashboard
```

arm64be:
Closes #19.

### DIFF
--- a/plugin/api/graph_data.py
+++ b/plugin/api/graph_data.py
@@ -17,7 +17,7 @@ log = logging.getLogger("theia-constellation")
 
 def _theia_home() -> Path:
     """Resolve $THEIA_HOME (default: ~/.hermes/) for the running dashboard."""
-    return Path(os.environ.get("THEIA_HOME", Path.home() / ".hermes"))
+    return Path(os.environ.get("THEIA_HOME") or os.environ.get("HERMES_HOME") or Path.home() / ".hermes")
 
 
 # ---------------------------------------------------------------------------
--- a/plugin/api/graph_data.py
+++ b/plugin/api/graph_data.py
@@ -18,7 +18,11 @@ log = logging.getLogger("theia-constellation")
 
 def _theia_home() -> Path:
     """Resolve data root: $THEIA_HOME, $HERMES_HOME, or ~/.hermes (by priority)."""
-    return Path(os.environ.get("THEIA_HOME") or os.environ.get("HERMES_HOME") or Path.home() / ".hermes")
+    return Path(
+        os.environ.get("THEIA_HOME")
+        or os.environ.get("HERMES_HOME")
+        or Path.home() / ".hermes"
+    )
 
 
 # ---------------------------------------------------------------------------
--- a/plugin/api/graph_data.py
+++ b/plugin/api/graph_data.py
@@ -2,7 +2,8 @@
 
 Locates and loads the pre-built graph JSON produced by ``theia-core``.
 The graph is generated offline (or via ``theia-core --watch``) and written
-to ``$THEIA_HOME/theia-graph.json``.  This module simply reads that file.
+to ``$THEIA_HOME/theia-graph.json`` (with ``$HERMES_HOME`` as fallback).
+This module simply reads that file.
 """
 
 from __future__ import annotations
@@ -16,14 +17,14 @@ log = logging.getLogger("theia-constellation")
 
 
 def _theia_home() -> Path:
-    """Resolve $THEIA_HOME (default: ~/.hermes/) for the running dashboard."""
+    """Resolve data root: $THEIA_HOME, $HERMES_HOME, or ~/.hermes (by priority)."""
     return Path(os.environ.get("THEIA_HOME") or os.environ.get("HERMES_HOME") or Path.home() / ".hermes")
 
 
 # ---------------------------------------------------------------------------
 # Search paths for the pre-built graph file, checked in order.
 #
-#   1. Default theia-core output:        $THEIA_HOME/theia-graph.json
+#   1. Default theia-core output:        $THEIA_HOME or $HERMES_HOME / theia-graph.json
 #   2. Plugin data dir (bundled fallback): <plugin>/data/graph.json
 #
 # The first path is the canonical location written by ``theia-core --watch``
--- a/theia-core/theia_core/__main__.py
+++ b/theia-core/theia_core/__main__.py
@@ -19,7 +19,7 @@ from theia_core.watch import watch_db
 
 
 def _theia_home() -> Path:
-    return Path(os.environ.get("THEIA_HOME", Path.home() / ".hermes"))
+    return Path(os.environ.get("THEIA_HOME") or os.environ.get("HERMES_HOME") or Path.home() / ".hermes")
 
 
 def _build(
--- a/theia-core/theia_core/__main__.py
+++ b/theia-core/theia_core/__main__.py
@@ -19,7 +19,9 @@ from theia_core.watch import watch_db
 
 
 def _theia_home() -> Path:
-    return Path(os.environ.get("THEIA_HOME") or os.environ.get("HERMES_HOME") or Path.home() / ".hermes")
+    return Path(
+        os.environ.get("THEIA_HOME") or os.environ.get("HERMES_HOME") or Path.home() / ".hermes"
+    )
 
 
 def _build(
--- a/README.md
+++ b/README.md
@@ -14,10 +14,12 @@ $HOME/.hermes/
   theia-graph.json    ← generated constellation graph
 ```
 
-Override the home directory with `THEIA_HOME`:
+Override the home directory with `THEIA_HOME` (or `HERMES_HOME` as fallback):
 
 ```bash
 export THEIA_HOME=/custom/path
+# or
+export HERMES_HOME=/custom/path
 ```
 
 See [`docs/superpowers/specs/2026-04-21-theia-design.md`](docs/superpowers/specs/2026-04-21-theia-design.md).
--- a/docs/integration.md
+++ b/docs/integration.md
@@ -2,7 +2,7 @@
 
 Both `theia-core` and `theia-panel` expect a `.hermes` directory (default
 `$HOME/.hermes`) containing session archives and the generated graph. Override
-with the `THEIA_HOME` environment variable.
+with `THEIA_HOME` (or `HERMES_HOME` as fallback).
 
 ## Mount surface
 
--- a/install.sh
+++ b/install.sh
@@ -30,7 +30,7 @@ set -euo pipefail
 # Config
 # ---------------------------------------------------------------------------
 REPO_URL="https://github.com/arm64be/theia"
-HERMES_HOME="${THEIA_HOME:-${HOME}/.hermes}"
+HERMES_HOME="${THEIA_HOME:-${HERMES_HOME:-${HOME}/.hermes}}"
 INSTALL_DIR="${HERMES_HOME}/hermes-theia"
 VENV_DIR="${INSTALL_DIR}/.venv"
 PLUGINS_DIR="${HERMES_HOME}/plugins"
--- a/theia-core/tests/test_cli.py
+++ b/theia-core/tests/test_cli.py
@@ -1,11 +1,14 @@
 import json
+import os
 import subprocess
 import sys
 from pathlib import Path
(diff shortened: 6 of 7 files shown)


zebster-cmd (commented):
# Code Review — PR #42: `fix(core): respect HERMES_HOME for profile-aware data paths`

## Summary
Two-line change adding `HERMES_HOME` as a secondary fallback in `_theia_home()` so Theia honors the active Hermes profile when `THEIA_HOME` is not explicitly set. Lookup order becomes `THEIA_HOME → HERMES_HOME → ~/.hermes`, matching the priority specified in issue #19. The fix is minimal, targeted, and addresses the reported bug.

## Correctness
- Logic is right: `os.environ.get("THEIA_HOME") or os.environ.get("HERMES_HOME") or Path.home() / ".hermes"` evaluates left-to-right; operator precedence is fine because `/` binds tighter than `or`.
- **Behavior change vs. issue's suggested patch**: PR uses `or`-chaining, while the issue suggested nested defaults. With `or`, an *empty* `THEIA_HOME=""` now silently falls through to `HERMES_HOME`; the previous `os.environ.get("THEIA_HOME", default)` returned `""` (causing path lookups to fail). Probably an improvement, but it is an undocumented semantic change — worth a comment or a short test.
- **Lazy vs. cached**: `theia-core/theia_core/__main__.py:102` snapshots `_theia_home()` once at startup; `plugin/api/graph_data.py:36-40` re-reads env on every request. Inconsistent but acceptable for the dashboard's long-running case (env doesn't change in-process).
- **Trailing slash / Windows**: `Path("/foo/")` normalizes correctly; `pathlib.Path` is OS-aware. No issues.
- **Inconsistency — other producers untouched**:
  - `theia-panel/vite.config.ts:5-6` still reads only `THEIA_HOME`. If a developer runs `vite serve` under an active profile, the panel's middleware will load the wrong `theia-graph.json`.
  - `install.sh:33` similarly ignores `HERMES_HOME` (less critical: install-time, but the variable is even *named* `HERMES_HOME` locally — confusing).
  - Docstrings/docs (`plugin/api/graph_data.py:5,19,26`, `plugin/api/plugin_api.py:6`, `docs/build-pipeline.md:36-37,61`, `README.md:17-20`) still describe only `THEIA_HOME`.

## Backward compatibility
Preserved. When neither env var is set, the result is `Path.home() / ".hermes"` exactly as before. Existing installs that set only `THEIA_HOME` are unaffected (it still wins).

## Security
Low risk: env vars come from the user's own shell. On shared / multi-tenant systems, however, anything that exports `HERMES_HOME` (e.g., a wrapper script, sudo-preserved env, or a CI runner inheriting env) now influences where Theia reads/writes. There is no `Path.resolve()` containment check, so `HERMES_HOME=/etc` would happily try to read `/etc/theia-graph.json`. Not a regression (same was true for `THEIA_HOME`), but the surface widens. Consider documenting that `HERMES_HOME` must be trusted, or validating that the resolved path exists and is a directory before use.

## Tests
None added. `theia-core/tests/test_cli.py` has no coverage for `_theia_home`. A minimal `monkeypatch.setenv`/`delenv` parametrized test asserting the three-tier priority (and empty-string behavior) would lock the contract cheaply.

## Suggestions
1. Update the docstring at `plugin/api/graph_data.py:19` and the comment at `plugin/api/graph_data.py:26` to reflect the new precedence.
2. Mirror the change in `theia-panel/vite.config.ts:5-6` for dev-server parity.
3. Add a unit test covering precedence and unset behavior.
4. Note the new env var in `README.md:17-20` and `docs/integration.md:5`.

## Verdict
**Approve with minor follow-ups.** The Python fix is correct, minimal, and backward-compatible; merge is safe. File a follow-up to (a) extend the same precedence to `vite.config.ts`, (b) update docs/docstrings, and (c) add a small regression test.

theia-core/theia_core/__main__.py
```diff
@@ -19,7 +19,7 @@
 
 
 def _theia_home() -> Path:
-    return Path(os.environ.get("THEIA_HOME", Path.home() / ".hermes"))
+    return Path(os.environ.get("THEIA_HOME") or os.environ.get("HERMES_HOME") or Path.home() / ".hermes")
```
zebster-cmd: Behavioural change worth calling out: previously `os.environ.get("THEIA_HOME", default)` returned `""` when the user explicitly set `THEIA_HOME=""` (and `Path("")` is `Path('.')`). With the new short-circuit `or` chain, an empty `THEIA_HOME` silently falls through to `HERMES_HOME` and then to `~/.hermes`. This is probably the right behaviour, but it's a semantic change — please document it inline so future readers don't accidentally revert it:

```suggestion
    # Precedence: THEIA_HOME > HERMES_HOME > ~/.hermes. Empty strings are treated as unset.
    return Path(os.environ.get("THEIA_HOME") or os.environ.get("HERMES_HOME") or Path.home() / ".hermes")
```

zebster-cmd: Follow-up parity work: this PR adds the `HERMES_HOME` fallback to `theia_core/__main__.py` and `plugin/api/graph_data.py`, but `theia-panel/vite.config.ts:5-6` and `install.sh:33` still read only `THEIA_HOME`. Users who set `HERMES_HOME` will get a working CLI + plugin but a dev server that points at the wrong directory. Worth a follow-up issue (or fold into this PR).

zebster-cmd (commented):
Two notes on the THEIA_HOME → HERMES_HOME fallback.

plugin/api/graph_data.py
```diff
@@ -17,7 +17,7 @@
 
 def _theia_home() -> Path:
     """Resolve $THEIA_HOME (default: ~/.hermes/) for the running dashboard."""
```
zebster-cmd: **WARNING**: Docstring is now stale — it still claims `Resolve $THEIA_HOME (default: ~/.hermes/)` with no mention of the HERMES_HOME fallback. Future readers (and users grepping for HERMES_HOME) won't discover the new resolution order from this function.

💡 """Resolve the dashboard data root.

Resolution order: $THEIA_HOME, then $HERMES_HOME, then ~/.hermes/.
Empty-string values are treated as unset.
"""

theia-core/theia_core/__main__.py
zebster-cmd: **WARNING**: `_theia_home()` is now defined identically in two places (plugin/api/graph_data.py and theia-core/theia_core/__main__.py). The original duplication was tolerable as a one-liner, but every future change to env var precedence (e.g. when issue #19's profile-awareness work lands) now has to be made twice and risks drift. Worth extracting before the next change touches it.

💡 Extract to theia_core/paths.py (or similar) and import from both call sites:

# theia_core/paths.py
def theia_home() -> Path:
    return Path(
        os.environ.get("THEIA_HOME")
        or os.environ.get("HERMES_HOME")
        or Path.home() / ".hermes"
    )

plugin/api/graph_data.py
```diff
@@ -17,7 +17,7 @@
...
-    return Path(os.environ.get("THEIA_HOME", Path.home() / ".hermes"))
+    return Path(os.environ.get("THEIA_HOME") or os.environ.get("HERMES_HOME") or Path.home() / ".hermes")
```
zebster-cmd: **SUGGESTION**: Behavioral note vs. the suggested fix in #19: `os.environ.get("THEIA_HOME", default)` returns `""` when the var is set-but-empty, which would then be passed to `Path()` and resolve to the CWD — a nasty footgun. The `or` chain in this PR correctly treats empty strings as unset. This is an improvement over the suggestion, not a regression, but it's worth calling out in the PR description so reviewers/users understand the intentional divergence.

zebster-cmd: **SUGGESTION**: No tests accompany this fix. Issue #19 is a regression-prone area (env var precedence, empty-string handling, profile-awareness coming next). A small parametrized test using `monkeypatch.setenv`/`delenv` covering the four cases (THEIA_HOME set, HERMES_HOME set, both set → THEIA_HOME wins, neither set → ~/.hermes, plus the empty-string case) would lock in the contract cheaply.

💡 @pytest.mark.parametrize("theia,hermes,expected", [
    ("/a", None, Path("/a")),
    (None, "/b", Path("/b")),
    ("/a", "/b", Path("/a")),
    ("",   "/b", Path("/b")),  # empty treated as unset
    (None, None, Path.home() / ".hermes"),
])
def test_theia_home(monkeypatch, theia, hermes, expected):
    for k, v in (("THEIA_HOME", theia), ("HERMES_HOME", hermes)):
        monkeypatch.delenv(k, raising=False)
        if v is not None:
            monkeypatch.setenv(k, v)
    assert _theia_home() == expected

zebster-cmd: **SUGGESTION**: Minor type-safety nit: the `or` chain has type `str | Path`, which `Path(...)` happily accepts but is slightly muddier than necessary. Normalizing to `str` (or branching) makes the intent clearer to readers and type-checkers.

💡 return Path(
    os.environ.get("THEIA_HOME")
    or os.environ.get("HERMES_HOME")
    or str(Path.home() / ".hermes")
)

zebster-cmd (commented):
## 🤖 Claude Code Review

The fix correctly resolves issue #19 by adding HERMES_HOME as a fallback. The `or` chain is actually slightly safer than the suggested nested `os.environ.get()` approach because it treats empty-string env vars as unset (a common gotcha). However, the change is duplicated across two files without extraction, the docstring is now stale, and there's no test coverage for the new fallback behavior.

*Reviewed by Claude Code via Hermes Agent*

PR #43, opened 2026-04-25
fix(panel): fix search bar z-index, left-edge clipping, and add smoot…

### LINKED ISSUE #39: bug(panel): search bar clips under filter bar and lacks transition when side panel opens
## Bug Report

Two related layout issues with the search bar when the side panel opens/closes:

1. **Search bar clips under the filter bar and into the left edge**
2. **Search bar snaps instantly while the side panel animates smoothly**

### Issue 1: Clipping and edge overflow

When the side panel opens, the search bar is shifted left via:

```css
.tp-panel-open .tp-search-bar {
  right: calc(min(420px, 45vw) + 12px);
  width: clamp(180px, calc(55vw - 36px), 320px);
}
```

**Problems:**

- **Z-index layering:** The filter bar has `z-index: 11`; the search bar has `z-index: 10`. When the panel opens and pushes the search bar leftward, it can overlap the filter bar horizontally. The search bar renders *behind* the filter bar, so its left portion becomes visually clipped/obscured.

- **Left-edge overflow:** On viewports around 700–800px wide, the shifted search bar's calculated `left` position can become negative or press against the container's left edge with no margin. The input text and dropdown get cut off.

  Example at 750px viewport:
  - Panel width: `min(420px, 45vw)` = 337px
  - Search bar right offset: 337 + 12 = 349px
  - Search bar width: `clamp(180px, calc(55vw - 36px), 320px)` = 55*7.5 - 36 ≈ 376px (clamped to 320px)
  - Search bar left edge: 750 - 349 - 320 = 81px → still positive, but with filter bar at left: 12px and max-width growing via wrap, the two can still collide.

  At 650px viewport:
  - Panel width: 292px
  - Search bar right offset: 304px
  - Search bar width: clamped to 320px
  - Search bar left edge: 650 - 304 - 320 = 26px → practically touching the left edge, no breathing room.

### Issue 2: Missing transition animation

The side panel has a smooth CSS transition:

```css
.tp-side-panel {
  transform: translateX(100%);
  transition: transform 220ms ease-out;
}
.tp-side-panel.tp-side-panel--open { transform: translateX(0); }
```

When a node is clicked, the panel slides in elegantly. However, the search bar has **no transition** on its `right` or `width` properties. It snaps instantly from `right: 12px` to `right: calc(min(420px, 45vw) + 12px)`, creating a jarring visual jump while the panel is still mid-animation.

### Repro

2. Resize the browser to ~750px wide.
4. **Observe:** The panel slides in smoothly. The search bar jumps instantly to its new position. If the filter bar has wrapped to 2+ rows, the search bar may render partially underneath it (lower z-index) or its left edge may be flush with the container boundary.
5. Close the panel (click × or click outside).
6. **Observe:** The panel slides out smoothly; the search bar snaps back instantly.

### Suggested fixes

**For clipping:**
1. Give the search bar `z-index: 12` (above filter bar) OR lower the filter bar to `z-index: 9` so the search bar sits on top when they overlap.
2. Add a `min-left` margin to the shifted search bar so it never gets closer than ~12px to the left edge:
   ```css
   .tp-panel-open .tp-search-bar {
     right: calc(min(420px, 45vw) + 12px);
     width: clamp(180px, calc(55vw - 36px), 320px);
     margin-left: 12px; /* prevents edge clipping */
   }
   ```
3. Or, on narrow viewports where collision is inevitable, hide the search bar earlier (e.g. `@media (max-width: 1000px)` instead of 900px).

**For missing transition:**
Add a CSS transition to `.tp-search-bar`:
```css
.tp-search-bar {
  transition: right 220ms ease-out, width 220ms ease-out;
}
```
This matches the panel's 220ms ease-out timing, so the search bar glides in sync with the panel.

### Environment

- File: `theia-panel/src/ui/layoutStyles.ts`

arm64be:
Closes #39.


zebster-cmd (commented):
# Review: PR #43 — fix(panel): search bar z-index, left-edge clipping, smooth transition

## Summary
Adds class `tp-search-bar`, switches the wrapper from left/translateX centering to a right-anchored equivalent, bumps z-index 10→12, adds `margin-left: 12px`, and injects a `transition: right/width 220ms ease-out` rule via a new `layoutStyles.ts`.

## Correctness
- **Issue #39's premise is not reproducible against `main`.** The issue references CSS classes (`.tp-panel-open`, `.tp-side-panel`, `.tp-search-bar`) and a "filter bar `z-index: 11`" that **do not exist** in this repo. `FilterBar.ts:67-76` sets no z-index; `SidePanel.ts:57-66` is class-less and styled via inline `cssText`; nothing toggles a `tp-panel-open` modifier on a parent. The PR fixes a state of the code that isn't in `main`.
- **Z-index hierarchy**: actual values are SearchBar 10 → 12 (`SearchBar.ts:30`), loading overlay 20 (`index.ts:317`), FilterBar/SidePanel/Tooltip none. There is no z:11 to invert against; the bump is cosmetic but does not invert anything (still below the loading overlay — fine).
- **Clipping is treated as symptom, not cause.** The "left edge clipping" only manifests if the bar is shifted left by a (nonexistent) `panel-open` rule. The added `margin-left: 12px` in `SearchBar.ts:32` actually **de-centers the bar 12 px right** — with `right: calc((100% - W)/2)` and explicit `width`, `margin-left` enters the absolute-positioning constraint and offsets the box. Net effect is asymmetric placement on every viewport, not just panel-open.
- **The right-anchored rewrite is mathematically identical** to the prior `left:50%; translateX(-50%)` (`SearchBar.ts:31`); it changes nothing on its own and only exists to make `right` "transition-able".

## Animation / UX
- **The transition is effectively dead code.** `layoutStyles.ts:9` transitions `right` and `width`, but those values are computed once in `applyWrapperStyle` and only re-set on `updateTheme`. Nothing in `index.ts`, `SidePanel.show/hide` (`SidePanel.ts:84-95, 187-190`), or elsewhere mutates the search bar's `right`/`width` when the side panel slides. The transition fires on theme changes (mostly imperceptible) and never on panel open/close.
- **`prefers-reduced-motion` is not respected.** The injected stylesheet has no `@media (prefers-reduced-motion: reduce) { .tp-search-bar { transition: none } }`. (Note: the pre-existing `SidePanel.ts:62` transition already had this gap; this PR is a missed opportunity to address both.)
- Timing/easing (220 ms ease-out) does match `SidePanel.ts:62` — good intent, no effect.

## Interaction with PR #36
PR #36 rewrites the same `applyWrapperStyle` to **left-anchored, full-width** (`left:12px; width:calc(100%-24px)`), which directly conflicts with this PR's right-anchored centering. Merge order matters: under PR #36 the search bar would sit on top of the filter bar (both at `left:12px`), making the z-index bump meaningful only after #36 lands — yet PR #43 was authored against the unmerged-and-different state.

## Code quality
- `layoutStyles.ts:1` uses module-level `let injected` for one-shot insertion. Other components (`SidePanel.ts:71-73`) inject styles inside their factory; mixing patterns. Per-instance `injectSearchBarStyles()` after `applyWrapperStyle()` (`SearchBar.ts:60`) is fine but the class isn't scoped — multiple panels share the rule (likely OK).
- New file for one rule is overkill; the rule could live next to `applyWrapperStyle` or in the existing focus stylesheet pattern.

## Suggestions
1. Close as not-applicable, or rebase after PR #36 with the actual conflict resolved.
2. If kept: remove `margin-left: 12px` (`SearchBar.ts:32`) — it visibly off-centers the bar.
3. Wire the transition to a real state change (e.g., `searchBar.setShifted(boolean)` called from `SidePanel.show/hide`) or drop it; transitioning unchanging properties is misleading.
4. Add `@media (prefers-reduced-motion: reduce)` guard around the new rule and, ideally, around `SidePanel.ts:62`.
5. Drop the new file; inline the rule.

## Verdict
**Request changes / do not merge.** The PR fixes a bug that isn't reproducible in `main`, ships a transition that never fires, and the `margin-left` change introduces a real off-center regression. Coordinate with PR #36 first and re-derive a fix from the actual on-`main` behavior.

theia-panel/src/ui/SearchBar.ts
```diff
@@ -26,8 +28,9 @@ export function createSearchBar(
 
   function applyWrapperStyle() {
     wrapper.style.cssText = `
-      position: absolute; top: 12px; left: 50%; transform: translateX(-50%);
-      z-index: 10; font: 13px/1.4 'Mondwest', var(--theia-font, ui-monospace, monospace);
+      position: absolute; top: 12px; right: calc((100% - min(320px, 50vw)) / 2); transform: none;
+      z-index: 12; margin-left: 12px;
```
zebster-cmd: `margin-left: 12px` here de-centres the bar relative to the carefully-computed `right: calc((100% - min(320px, 50vw)) / 2)` on the previous line. Together they push the search bar 12 px past the visual centre. Drop the margin and let the `right: calc(...)` math do the centering on its own:

```suggestion
      z-index: 12;
```

theia-panel/src/ui/layoutStyles.ts
```diff
@@ -0,0 +1,13 @@
+let injected = false;
+
+export function injectSearchBarStyles(): void {
+  if (injected) return;
+  injected = true;
+  const style = document.createElement("style");
+  style.textContent = `
+    .tp-search-bar {
+      transition: right 220ms ease-out, width 220ms ease-out;
```
zebster-cmd: This transition never actually fires. `right` and `width` on `.tp-search-bar` are only ever (re-)set inside `applyWrapperStyle()`, which runs at construction and on theme change — both of which write the *full* cssText in one shot, so there's no value-to-value interpolation. To get the intended slide-when-side-panel-opens animation, either: (a) expose a `setShifted(open: boolean)` method on `SearchBar` that toggles a class which overrides `right`, and call it from `SidePanel.show/hide`; or (b) drop this rule and the surrounding plumbing entirely until it's wired up.

zebster-cmd: Also: even once it does fire, please respect users' motion preferences. Wrap it in a `prefers-reduced-motion` guard so the search bar doesn't slide for users who've opted out:

```css
@media (prefers-reduced-motion: reduce) {
  .tp-search-bar { transition: none; }
}
```

zebster-cmd (commented):
Three layout / motion notes.

PR #44, opened 2026-04-25
fix(panel): use dashboard font stack instead of hardcoded Mondwest — …

### LINKED ISSUE #32: bug(panel): font does not match the rest of the Hermes dashboard
## Bug Report

The constellation panel uses a different font family than the rest of the Hermes dashboard (sessions list, sidebar, config, etc.), creating a visual inconsistency.

### Root cause analysis

The dashboard uses a sans-serif system font stack:

```css
/* web/src/index.css */
html { font-family: var(--theme-font-sans); }
/* → system-ui, -apple-system, "Segoe UI", Roboto, "Helvetica Neue", Arial, sans-serif */
```

The plugin correctly extracts this via `getComputedStyle(document.documentElement).fontFamily` and passes it to the iframe as a query param. However, the panel's CSS hardcodes a different font stack.

**Panel font stack (current):**

```css
/* SearchBar.ts */
font: 13px/1.4 'Mondwest', var(--theia-font, ui-monospace, monospace);

/* FilterBar.ts */
font: 10px/1.4 'Mondwest', var(--theia-font, ui-monospace, monospace);

/* Tooltip.ts */
font: 12px/1.4 'Mondwest', var(--theia-font, ui-monospace, monospace);

/* SidePanel.ts */
font: 13px/1.6 var(--theia-font, ui-monospace, monospace);
```

**Problems:**

1. **`'Mondwest'` is placed first** in SearchBar, FilterBar, and Tooltip. If the browser has this font available (or loads it), it overrides the dashboard's font entirely. Mondwest is a decorative/display font that looks nothing like `system-ui` or `Inter`.

2. **The fallback is `ui-monospace, monospace`** — a monospace stack. If `--theia-font` is not set (e.g. standalone use, dev mode without query params, or the dashboard's font token fails to propagate), the panel renders everything in a code font instead of a readable sans-serif.

3. **Inconsistent usage across components** — SidePanel does not hardcode Mondwest; SearchBar/FilterBar/Tooltip do. So even within the panel, different UI elements use different fonts.

### Repro

1. Open the Constellation tab in the dashboard.
2. Compare the font of the search bar, filter toggles, and tooltip with the font in the sessions list or sidebar.
3. The panel elements use a narrower, more decorative/monospace font while the dashboard uses the system sans-serif.

### Suggested fix

**Option A (preferred):** Remove `'Mondwest'` from all hardcoded font stacks and use `--theia-font` as the primary, with a sensible sans-serif fallback:

```css
font: 13px/1.4 var(--theia-font, system-ui, -apple-system, sans-serif);
```

Update these files:
- `theia-panel/src/ui/SearchBar.ts`
- `theia-panel/src/ui/FilterBar.ts`
- `theia-panel/src/ui/Tooltip.ts`
- `theia-panel/src/ui/SidePanel.ts`
- `theia-panel/src/ui/layoutStyles.ts` (if present)

Also update the default in `Theme.ts`:

```ts
font: "system-ui, -apple-system, sans-serif",
```

**Option B:** If Mondwest is an intentional aesthetic choice for the constellation "space theme," load it explicitly as a web font and make it opt-in (e.g. via a `?font=Mondwest` query param) rather than overriding the dashboard font by default.

### Environment

- Files: `theia-panel/src/ui/SearchBar.ts`, `FilterBar.ts`, `Tooltip.ts`, `SidePanel.ts`, `Theme.ts`

arm64be:
Closes #32.

### DIFF
--- a/theia-panel/src/ui/SearchBar.ts
+++ b/theia-panel/src/ui/SearchBar.ts
@@ -27,7 +27,7 @@ export function createSearchBar(
   function applyWrapperStyle() {
     wrapper.style.cssText = `
       position: absolute; top: 12px; left: 50%; transform: translateX(-50%);
-      z-index: 10; font: 13px/1.4 'Mondwest', var(--theia-font, ui-monospace, monospace);
+      z-index: 10;       font: 13px/1.4 var(--theia-font, system-ui, -apple-system, sans-serif);
       color: #${theme.fg}; width: min(320px, 50vw);
     `;
   }
--- a/theia-panel/src/ui/SearchBar.ts
+++ b/theia-panel/src/ui/SearchBar.ts
@@ -1,6 +1,6 @@
 import type { TheiaGraph } from "../data/types";
 import type { ThemeTokens } from "./Theme";
-import { themeBgAlpha } from "./Theme";
+import { themeBgAlpha, FONT_STACK } from "./Theme";
 import { escape } from "./utils";
 
 export interface SearchResult {
@@ -27,7 +27,7 @@ export function createSearchBar(
   function applyWrapperStyle() {
     wrapper.style.cssText = `
       position: absolute; top: 12px; left: 50%; transform: translateX(-50%);
-      z-index: 10;       font: 13px/1.4 var(--theia-font, system-ui, -apple-system, sans-serif);
+      z-index: 10; font: 13px/1.4 var(--theia-font, ${FONT_STACK});
       color: #${theme.fg}; width: min(320px, 50vw);
     `;
   }
--- a/plugin/src/index.js
+++ b/plugin/src/index.js
@@ -91,7 +91,7 @@
     var accent   = toHex(cssVar("--color-warning") || cssVar("--midground-base"), "ffc477");
     var border   = toHex(cssVar("--color-border"), "ffffff26");
     var font     = root.getPropertyValue("font-family").trim()
-      || "'Mondwest', ui-monospace, monospace";
+      || "system-ui, -apple-system, 'Segoe UI', Roboto, 'Helvetica Neue', Arial, sans-serif";
     return { bg, fg, fg2, midground, accent, border, font, radius: "0" };
   }
 
--- a/theia-panel/src/index.ts
+++ b/theia-panel/src/index.ts
@@ -11,7 +11,7 @@ import { createTooltip } from "./ui/Tooltip";
 import { createFilterBar } from "./ui/FilterBar";
 import { createSearchBar } from "./ui/SearchBar";
 import { createSidePanel } from "./ui/SidePanel";
-import { readTheme, applyTheme, onThemeMessage } from "./ui/Theme";
+import { readTheme, applyTheme, onThemeMessage, FONT_STACK } from "./ui/Theme";
 import type { ThemeTokens } from "./ui/Theme";
 
 export interface PanelOptions {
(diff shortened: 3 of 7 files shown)


zebster-cmd (commented):
# Code Review — PR #44 (`fix/issue-32-font-mismatch`)

## Summary
Removes hardcoded `'Mondwest'` from panel font stacks and switches the fallback from monospace (`ui-monospace, monospace`) to a sans-serif system stack, aligning with the dashboard per issue #32 (Option A). 5 files touched, +6/−6 LOC.

## Correctness — single source of truth
The propagation chain is intact: dashboard `--theme-font-sans` → `plugin/src/index.js:93-94` (`getComputedStyle().font-family`) → query param → `theia-panel/src/ui/Theme.ts:52` → `--theia-font` CSS var (`Theme.ts:67`). The panel correctly defers to `var(--theia-font, …)` as the primary source.

**However, two `'Mondwest'` references survive the PR and contradict the stated goal:**
- `theia-panel/src/index.ts:316` — `createLoadingOverlay` still uses `font: 13px/1.4 'Mondwest', var(--theia-font, ui-monospace, monospace);` (Mondwest ahead of the var, plus monospace fallback).
- `plugin/src/index.js:94` — default literal `"'Mondwest', ui-monospace, monospace"` is still emitted into the query string when the dashboard's computed `font-family` is empty (e.g. standalone iframe, dev mode). This re-introduces exactly the bug #32 describes.

The issue body listed only the four UI files, so these omissions are understandable, but they leave the fix incomplete.

## Fallback chain
New stack `system-ui, -apple-system, sans-serif` (e.g. `Theme.ts:38`, `SidePanel.ts:61`) is reasonable but shorter than the dashboard's stack quoted in #32 (adds `"Segoe UI", Roboto, "Helvetica Neue", Arial`). Only used when `--theia-font` is unset, so low risk, but worth matching for parity.

## Code quality
- `SearchBar.ts:30` has a whitespace artefact: `z-index: 10;       font: …` (multiple spaces left after the edit). Cosmetic only inside a template literal, but sloppy.
- The fallback string `var(--theia-font, system-ui, -apple-system, sans-serif)` is duplicated across 5 sites. Extract once (e.g. `FONT_STACK` exported from `Theme.ts`) to make this kind of drift impossible next time — that *is* the SSOT the PR title promises.

## Risk
Low. Mondwest was unintentional/inconsistent (SidePanel never had it). No web font is loaded for it, so removal cannot regress branding. Theme override via query param still works.

## Suggestions
1. Also patch `theia-panel/src/index.ts:316` and `plugin/src/index.js:94`.
2. Define `export const FONT_STACK` in `Theme.ts` and reuse in all `font:` declarations.
3. Match the dashboard's full sans stack in the fallback.
4. Clean the double-space in `SearchBar.ts:30`.

## Verdict
**Request changes.** The fix is directionally correct but two surviving `'Mondwest'` sites mean issue #32's repro can still trigger (loading overlay, and standalone/dev mode via the plugin fallback).

zebster-cmd (commented):
One whitespace nit — the other two anchors I'd raise (`theia-panel/src/index.ts:316` createLoadingOverlay, and the `plugin/src/index.js:94` fallback) aren't part of this PR's diff, so flagging here only.

theia-panel/src/ui/SearchBar.ts
```diff
@@ -27,7 +27,7 @@ export function createSearchBar(
   function applyWrapperStyle() {
     wrapper.style.cssText = `
       position: absolute; top: 12px; left: 50%; transform: translateX(-50%);
-      z-index: 10; font: 13px/1.4 'Mondwest', var(--theia-font, ui-monospace, monospace);
+      z-index: 10;       font: 13px/1.4 var(--theia-font, system-ui, -apple-system, sans-serif);
```
zebster-cmd: Stray run of spaces between `;` and `font:` — looks like a leftover from removing `'Mondwest', `. Collapse to a single space:

```suggestion
      z-index: 10; font: 13px/1.4 var(--theia-font, system-ui, -apple-system, sans-serif);
```

Separate follow-up notes (anchors not in this diff): `theia-panel/src/index.ts:316` (createLoadingOverlay) still hard-codes `'Mondwest', ` ahead of the `var(--theia-font, ...)` fallback chain, and `plugin/src/index.js:94`'s default for the dashboard's computed-font fallback is still `'Mondwest', ui-monospace, monospace`. Both should switch to `system-ui, -apple-system, sans-serif` so the dashboard renders consistently when Mondwest isn't installed.

PR #47, opened 2026-04-25
feat(core): add cron-chain edges to link cron-spawned sessions in constellation

### LINKED ISSUE #46: feat(core): add cron-chain edges to link cron-spawned sessions in constellation
## Feature Request

Add graph edges that connect cron-spawned sessions to each other, so users can visually trace recurring task lineages in the constellation.

### Current state

The graph already supports `subagent` edges (parent → child) for delegated tasks. Cron jobs also spawn sessions, but those sessions appear as isolated nodes with no visual connection to other runs of the same cron job.

### Desired behaviour

When multiple sessions share the same `cron_job_id` (or similar cron identifier), draw edges between them so the constellation shows:

- Which sessions were spawned by cron jobs
- The chronological chain of cron executions for the same job
- Dependencies or causality between scheduled runs

### Proposed edge type

`cron-chain` — connects consecutive runs of the same cron job:

```
cron-job-A-run-1  →  cron-job-A-run-2  →  cron-job-A-run-3
```

Or `cron-sibling` — connects all runs of the same cron job into a cluster:

```
cron-job-A-run-1  —  cron-job-A-run-2
        |                 |
cron-job-A-run-3  —  cron-job-A-run-4
```

### Open questions

1. **Column availability** — Does `state.db` store `cron_job_id` or similar? The `sessions` table has `source` (set to `"cron"` for cron sessions) but may not have a cron job identifier column.

2. **Edge semantics** — Should the edge represent:
   - Temporal adjacency (run N → run N+1)?
   - Shared origin (all runs of the same job are siblings)?
   - Both (temporal chain + a shared parent node for the cron job itself)?

3. **Self-loops** — If a cron job spawns a subagent, should the subagent session link to the cron parent via `subagent` AND link to other cron runs via `cron-chain`?

### Suggested implementation

1. Add `cron_job_id` detection to `theia-core/theia_core/ingest.py` (or read from `state.db` if Hermes already stores it).
2. Add a new detector `theia-core/theia_core/detect/cron.py` that groups sessions by cron job ID and creates edges.
3. Add `cron-chain` to the edge kind enum in `schemas/graph.schema.json`.
4. Style the edges differently from `subagent` (e.g. dashed line, different colour) so users can distinguish scheduled runs from delegated tasks.

### Acceptance criteria

- [ ] Cron-spawned sessions are visually linked in the constellation
- [ ] Edge type is documented in `schemas/graph.schema.json`
- [ ] Filter bar toggle for `cron-chain` edges
- [ ] Distinct visual style (colour/line pattern) from `subagent` edges
- [ ] Works alongside existing `subagent` edges (no conflict)

### Environment

- Branch: `main`

arm64be:
## Summary

- Adds `cron-chain` edge type to link consecutive runs of the same cron job in the constellation graph
- New detector `theia-core/theia_core/detect/cron.py` groups sessions by `cron_job_id` and creates edges between runs
- Updates graph schema, frontend types, styling, filter bar toggle, and simulation physics

Closes #46

### DIFF
--- a/theia-core/theia_core/detect/cron.py
+++ b/theia-core/theia_core/detect/cron.py
@@ -21,7 +21,7 @@ def detect_cron_chain(sessions: list[Session]) -> list[Edge]:
     edges: list[Edge] = []
     for job_id, group in by_job.items():
         group.sort(key=lambda s: s.started_at)
-        for prev_sess, next_sess in zip(group, group[1:], strict=True):
+        for prev_sess, next_sess in zip(group, group[1:]):
             hours = (next_sess.started_at - prev_sess.started_at).total_seconds() / 3600
             weight = max(0.1, min(1.0, 1.0 / (1.0 + hours * 0.1)))
             edges.append(
--- a/theia-core/theia_core/detect/cron.py
+++ b/theia-core/theia_core/detect/cron.py
@@ -1,0 +1,39 @@
+from __future__ import annotations
+
+from collections import defaultdict
+
+from theia_core.detect import Edge
+from theia_core.ingest import Session
+
+
+def detect_cron_chain(sessions: list[Session]) -> list[Edge]:
+    """Create edges between consecutive runs of the same cron job.
+
+    Sessions spawned by the same cron job are grouped by ``cron_job_id``,
+    sorted chronologically, and an edge is created between each consecutive
+    pair.  This lets users visually trace recurring task lineages.
+    """
+    by_job: dict[str, list[Session]] = defaultdict(list)
+    for sess in sessions:
+        if sess.cron_job_id:
+            by_job[sess.cron_job_id].append(sess)
+
+    edges: list[Edge] = []
+    for job_id, group in by_job.items():
+        group.sort(key=lambda s: s.started_at)
+        for prev_sess, next_sess in zip(group, group[1:], strict=True):
+            hours = (next_sess.started_at - prev_sess.started_at).total_seconds() / 3600
+            weight = max(0.1, min(1.0, 1.0 / (1.0 + hours * 0.1)))
+            edges.append(
+                Edge(
+                    source=prev_sess.id,
+                    target=next_sess.id,
+                    kind="cron-chain",
+                    weight=weight,
+                    evidence={
+                        "cron_job_id": job_id,
+                        "interval_hours": round(hours, 2),
+                    },
+                )
+            )
+    return edges
--- a/theia-core/theia_core/detect/cron.py
+++ b/theia-core/theia_core/detect/cron.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 from collections import defaultdict
+from itertools import pairwise
 
 from theia_core.detect import Edge
 from theia_core.ingest import Session
@@ -21,7 +22,7 @@ def detect_cron_chain(sessions: list[Session]) -> list[Edge]:
     edges: list[Edge] = []
     for job_id, group in by_job.items():
         group.sort(key=lambda s: s.started_at)
-        for prev_sess, next_sess in zip(group, group[1:]):
+        for prev_sess, next_sess in pairwise(group):
             hours = (next_sess.started_at - prev_sess.started_at).total_seconds() / 3600
             weight = max(0.1, min(1.0, 1.0 / (1.0 + hours * 0.1)))
             edges.append(
--- a/theia-panel/src/index.ts
+++ b/theia-panel/src/index.ts
@@ -132,7 +132,7 @@ export async function mount(
     modelFilter?: string | null,
   ): Set<string> {
     const kindVisible = new Set<string>();
-    if (enabledKinds.has("subagent")) {
+    if (enabledKinds.has("subagent") || enabledKinds.has("cron-chain")) {
       for (const node of graph.nodes) {
         kindVisible.add(node.id);
       }
--- a/theia-panel/src/index.ts
+++ b/theia-panel/src/index.ts
@@ -132,7 +132,7 @@ export async function mount(
     modelFilter?: string | null,
   ): Set<string> {
     const kindVisible = new Set<string>();
-        if (enabledKinds.has("subagent")) {
+    if (enabledKinds.has("subagent")) {
       for (const node of graph.nodes) {
         kindVisible.add(node.id);
       }
--- a/theia-panel/src/index.ts
+++ b/theia-panel/src/index.ts
@@ -132,7 +132,7 @@ export async function mount(
     modelFilter?: string | null,
   ): Set<string> {
     const kindVisible = new Set<string>();
-    if (enabledKinds.has("subagent") || enabledKinds.has("cron-chain")) {
+        if (enabledKinds.has("subagent")) {
       for (const node of graph.nodes) {
         kindVisible.add(node.id);
       }
--- a/theia-panel/src/ui/SidePanel.ts
+++ b/theia-panel/src/ui/SidePanel.ts
@@ -50,7 +50,7 @@ export function createSidePanel(
     onNavigate?: (nodeId: string) => void;
     onClose?: () => void;
     onFocusToggle?: (enabled: boolean) => void;
-  }
+  },
 ) {
   let theme = initialTheme;
   const el = document.createElement("aside");
--- a/theia-panel/src/ui/SidePanel.ts
+++ b/theia-panel/src/ui/SidePanel.ts
@@ -280,6 +280,13 @@ function renderEdge(
     detail = `<div style="${detailAttrStyle(theme)}">
       ${label}: <span style="color:#${theme.accent}">${escape(String(displayId ?? "?"))}</span>
     </div>`;
+  } else if (e.kind === "cron-chain") {
+    const jobId = ev.cron_job_id;
+    const intervalHours = ev.interval_hours;
+    detail = `<div style="${detailAttrStyle(theme)}">
+      cron job: <span style="color:#${theme.accent}">${escape(String(jobId ?? "?"))}</span>
+      ${intervalHours !== undefined ? `\u00b7 interval ${Number(intervalHours).toFixed(1)}h` : ""}
+    </div>`;
   }
 
   return `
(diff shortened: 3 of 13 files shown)


theia-core/theia_core/detect/cron.py
```diff
@@ -0,0 +1,39 @@
...
+    """
+    by_job: dict[str, list[Session]] = defaultdict(list)
+    for sess in sessions:
+        if sess.cron_job_id:
+            by_job[sess.cron_job_id].append(sess)
+
+    edges: list[Edge] = []
+    for job_id, group in by_job.items():
+        group.sort(key=lambda s: s.started_at)
+        for prev_sess, next_sess in zip(group, group[1:], strict=True):
```
zebster-cmd: **CRITICAL**: `zip(group, group[1:], strict=True)` always raises ValueError when len(group) >= 2 — `group` and `group[1:]` differ in length by 1, and `strict=True` enforces equal lengths. This means detect_cron_chain() will crash on every real cron job (any group with at least 2 sessions yields exactly the case `strict=True` is designed to reject). Result: the entire build pipeline blows up the moment a cron session repeats.

💡 Drop `strict=True` (i.e. `zip(group, group[1:])`) or, more idiomatically, use `from itertools import pairwise` and iterate `for prev_sess, next_sess in pairwise(group):`. Either preserves the intended pairwise semantics without the length-mismatch crash.

theia-panel/src/index.ts
```diff
@@ -132,7 +132,7 @@ export async function mount(
     modelFilter?: string | null,
   ): Set<string> {
     const kindVisible = new Set<string>();
-    if (enabledKinds.has("subagent")) {
+    if (enabledKinds.has("subagent") || enabledKinds.has("cron-chain")) {
```
zebster-cmd: **WARNING**: Adding `enabledKinds.has("cron-chain")` to the "reveal every node" branch is over-broad. The subagent special case exists because subagent edges encode parent/child relationships across all sessions; cron-chain edges only ever connect sessions whose id starts with `cron_…`. With this change, toggling cron-chain alone will display every node in the graph, defeating the filter. Cron-chain should follow the normal edge-driven visibility path used by memory-share / cross-search / tool-overlap.

💡 Remove the `|| enabledKinds.has("cron-chain")` clause and let cron-chain edges drive `hasNonSubagentConnection` through the existing else-branch loop, so only sessions actually participating in a cron chain become visible.

theia-panel/src/ui/SidePanel.ts
```diff
@@ -280,6 +280,13 @@ function renderEdge(
     detail = `<div style="${detailAttrStyle(theme)}">
       ${label}: <span style="color:#${theme.accent}">${escape(String(displayId ?? "?"))}</span>
     </div>`;
+  } else if (e.kind === "cron-chain") {
+    const jobId = ev.cron_job_id;
+    const intervalHours = ev.interval_hours;
+    detail = `<div style="${detailAttrStyle(theme)}">
+      cron job: <span style="color:#${theme.accent}">${escape(String(jobId ?? "?"))}</span>
+      ${intervalHours !== undefined ? `\u00b7 interval ${Number(intervalHours).toFixed(1)}h` : ""}
```
zebster-cmd, line 288: **SUGGESTION**: `ev.interval_hours` is typed as `unknown`. The `!== undefined` guard lets through any non-number value (string, null, object), and `Number(value).toFixed(1)` then renders the literal text "NaN.0"… actually "NaN" — i.e. UI noise on malformed evidence. Other branches in this file already gate on `typeof === 'number'` style checks (e.g. jaccard guarded by Array.isArray for sharedTools).

💡 Narrow the type before formatting, e.g. `const intervalHours = typeof ev.interval_hours === "number" ? ev.interval_hours : undefined;` and keep the `!== undefined` guard. That removes the NaN risk and matches the surrounding pattern.

theia-core/theia_core/detect/cron.py
```diff
@@ -0,0 +1,39 @@
...
+def detect_cron_chain(sessions: list[Session]) -> list[Edge]:
+    """Create edges between consecutive runs of the same cron job.
+
+    Sessions spawned by the same cron job are grouped by ``cron_job_id``,
+    sorted chronologically, and an edge is created between each consecutive
+    pair.  This lets users visually trace recurring task lineages.
+    """
+    by_job: dict[str, list[Session]] = defaultdict(list)
+    for sess in sessions:
+        if sess.cron_job_id:
```
zebster-cmd, line 19: **SUGGESTION**: Minor: `_extract_cron_job_id` returns the empty string for an id like `cron__20260330_171037` (split yields `["cron", "", "20260330_171037"]`). Empty string is falsy, so the downstream `if sess.cron_job_id` filters it out, but it's a tiny footgun if anyone iterates `by_job` directly later.

💡 Add `if not parts[1]: return None` after the length check, or `return parts[1] or None`.

zebster-cmd (changes requested):
## Code Review

Adds a "cron-chain" edge kind end-to-end (schema, Python detector, ingest, types, physics, palette, FilterBar, SidePanel). The plumbing is consistent and well-typed, but the detector contains a critical bug: `zip(group, group[1:], strict=True)` raises `ValueError` for every cron job with ≥2 sessions, so the feature will throw at runtime on any real input. There is also a visibility-filter regression where enabling only "cron-chain" reveals every node in the graph.

zebster-cmd:
## Code Review Summary

**Verdict:** REQUEST_CHANGES

Adds a "cron-chain" edge kind end-to-end (schema, Python detector, ingest, types, physics, palette, FilterBar, SidePanel). The plumbing is consistent and well-typed, but the detector contains a critical bug: `zip(group, group[1:], strict=True)` raises `ValueError` for every cron job with ≥2 sessions, so the feature will throw at runtime on any real input. There is also a visibility-filter regression where enabling only "cron-chain" reveals every node in the graph.

### Findings
- 🔴 **theia-core/theia_core/detect/cron.py:24** — `zip(group, group[1:], strict=True)` always raises ValueError when len(group) >= 2 — `group` and `group[1:]` differ in length by 1, and `strict=True` enforces equal lengths. This means detect_cron_chain() will crash on every real cron job (any group with at least 2 sessions yields exactly the case `strict=True` is designed to reject). Result: the entire build pipeline blows up the moment a cron session repeats.
- ⚠️ **theia-panel/src/index.ts:135** — Adding `enabledKinds.has("cron-chain")` to the "reveal every node" branch is over-broad. The subagent special case exists because subagent edges encode parent/child relationships across all sessions; cron-chain edges only ever connect sessions whose id starts with `cron_…`. With this change, toggling cron-chain alone will display every node in the graph, defeating the filter. Cron-chain should follow the normal edge-driven visibility path used by memory-share / cross-search / tool-overlap.
- 💡 **theia-panel/src/ui/SidePanel.ts:288** — `ev.interval_hours` is typed as `unknown`. The `!== undefined` guard lets through any non-number value (string, null, object), and `Number(value).toFixed(1)` then renders the literal text "NaN.0"… actually "NaN" — i.e. UI noise on malformed evidence. Other branches in this file already gate on `typeof === 'number'` style checks (e.g. jaccard guarded by Array.isArray for sharedTools).
- 💡 **theia-core/theia_core/detect/cron.py:18** — Minor: `_extract_cron_job_id` returns the empty string for an id like `cron__20260330_171037` (split yields `["cron", "", "20260330_171037"]`). Empty string is falsy, so the downstream `if sess.cron_job_id` filters it out, but it's a tiny footgun if anyone iterates `by_job` directly later.

zebster-cmd:
Closed as effectively merged via PR #54 — partially literal, partially via parallel commits.

**Literal merge via #54:** The cron-chain feature commit `4690069` from this branch was carried into `main` as part of PR #54's merge commit `0f74cb7`. This was unintentional — `4690069` should have shipped through this PR (#47), not via cross-contamination on #54's branch — but the code itself is the same. Process retrospective at the bottom of this comment.

**Content-equivalent on main via parallel commits:** The two follow-up commits on this branch (`f7aaae4`, `7f68787`) are not literally on `main`, but every line they propose has an equivalent or improved counterpart on `main`, delivered by independent commits authored during PR #54's review cycle:

| This PR's change | Equivalent on `main` |
|---|---|
| `cron.py`: drop `strict=True` from `zip` | `f122d66` — uses `itertools.pairwise` instead (modern idiom, also ruff B905 compliant) |
| `cron.py`: add `from itertools import pairwise` + `pairwise(group)` | `f122d66` — identical |
| `index.ts:135`: remove `cron-chain` from forced-visibility check | `fd8e28d` — identical |
| `index.ts`: 4-space indent fix | identical on `main` |
| `package.json`: chain `prettier --write src/data/types.ts` into `generate-types` | `e91cbc0` — identical |
| `types.ts`: multi-line `edges.kind` union + multi-line `evidence` | `0cfad35` — identical |
| `SidePanel.ts:53`: trailing comma after parameter `}` | `0cfad35` — identical |

**Plus** `main` carries three additional review-fix commits that this PR doesn't have:
- `96037c2` — `_extract_cron_job_id` corrected to `rsplit("_", 2)` with shape check
- `37f9a23` — `interval_hours` rendering hardened with `typeof === "number" && Number.isFinite` guard
- `180ee00` — extracted `_FALLOFF_PER_HOUR` / `_MIN_WEIGHT` named constants with falloff comment

Net effect: every line on this branch is on `main`, and `main` carries strict improvements beyond what this PR proposed. Merging this branch tip today would be either a no-op or a regression (`pairwise` → `zip`). GitHub will mark this as `Closed` rather than `Merged` because the branch tip SHA (`7f68787`) isn't reachable from `main` — but the work itself shipped.

**Process retrospective.** This branch and PR #54's branch shared a commit ancestry (`4690069 → 6ef4e49 → 9dfcc82 → dd31ff2`) instead of branching independently from `main`. PR #54 was titled "fix: replace tarball download with git clone in install.sh" but its branch carried the cron-chain feature plus the pitch-rotation fix at its base, so merging #54 unintentionally landed both. Going forward, any PR review on this repo should start with `git log <base>..<branch>` checked against the PR's stated scope, and feature branches should be created off `main` directly to prevent this.

PR #48, opened 2026-04-25
fix(api): derive dev panel URL from request host, add THEIA_DEV_HOST override

### LINKED ISSUE #45: bug(api): dev panel URL hardcoded to localhost:5173 — breaks in WSL/Docker/remote
## Bug Report

The development mode panel URL is hardcoded to `http://localhost:5173`, which breaks in any environment where the browser and the Hermes server are not on the same machine (WSL, Docker, remote servers, CI).

### Root cause

`plugin/api/plugin_api.py` hardcodes the dev server host:

```python
THEIA_DEV_PORT = os.environ.get("THEIA_DEV_PORT", "5173")

if THEIA_ENV == "development":
    config["dev_panel_url"] = f"http://localhost:{THEIA_DEV_PORT}"
```

**Why this fails in live/shared environments:**

1. **WSL with Windows browser** — The Vite dev server binds to `127.0.0.1:5173` inside WSL. Windows-side `localhost:5173` does not reach the WSL loopback unless `netsh interface portproxy` is set up.

2. **Docker / containerized deployments** — `localhost:5173` inside the container is not accessible from the host browser unless the port is explicitly published (`-p 5173:5173`).

3. **Remote servers / cloud instances** — The browser is on the user's laptop; `localhost:5173` resolves to the laptop, not the server.

4. **Multiple developers on shared machine** — If 5173 is already taken by another dev's Vite instance, there's no auto-assignment or conflict detection.

### Current hardcoded references to 5173

| File | Hardcoded value | Context |
|---|---|---|
| `plugin/api/plugin_api.py:45` | `"5173"` | `THEIA_DEV_PORT` default |
| `plugin/api/plugin_api.py:62` | `"http://localhost:{THEIA_DEV_PORT}"` | dev_panel_url |
| `Makefile:26` | `DEV_PORT := 5173` | Vite dev server port |
| `Makefile:72` | `--port $(DEV_PORT)` | `make dev` target |
| `examples/dashboard.html:361` | `"http://localhost:5173"` | demo iframe src |
| `docs/build-pipeline.md:57` | `http://localhost:5173` | documentation |

### Suggested fixes

**Option A: Derive host from the incoming request**

```python
from fastapi import Request

@router.get("/config")
async def get_config(request: Request):
    config = {"env": THEIA_ENV, "version": "0.1.0"}
    if THEIA_ENV == "development":
        host = request.headers.get("x-forwarded-host") or request.headers.get("host", "localhost")
        host = host.split(":")[0]
        config["dev_panel_url"] = f"http://{host}:{THEIA_DEV_PORT}"
    return config
```

**Option B: Explicit `THEIA_DEV_HOST` env var**

```python
THEIA_DEV_HOST = os.environ.get("THEIA_DEV_HOST", "localhost")
config["dev_panel_url"] = f"http://{THEIA_DEV_HOST}:{THEIA_DEV_PORT}"
```

**Recommended:** Option A + Option B as fallback.

### Acceptance criteria

- [ ] `dev_panel_url` is no longer hardcoded to `localhost`
- [ ] Works out of the box in WSL, Docker, and remote deployments
- [ ] Port 5173 can be overridden via `THEIA_DEV_PORT` without editing source
- [ ] Add port validation (reject < 1024 and well-known Hermes ports like 9119)
- [ ] Update `Makefile` `make dev` to bind to `0.0.0.0` by default
- [ ] Update documentation

### Environment

- Branch: `main`
- Files: `plugin/api/plugin_api.py`, `Makefile`, `examples/dashboard.html`

arm64be:
## Summary

- Derives `dev_panel_url` from `THEIA_DEV_HOST` env var or defaults to `localhost` (request headers are never trusted)
- Adds `THEIA_DEV_HOST` env var for explicit override
- Adds port validation (rejects <1024, >65535, and port 9119)
- `make dev` binds to `127.0.0.1` by default
- Updates dashboard example and documentation

Closes #45

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -69,10 +69,10 @@ dev: dev-link ## Start dev environment (Vite hot-reload + plugin symlink)
 	@echo "    Panel:     http://localhost:$(DEV_PORT)"
 	@echo "    Dashboard: run 'THEIA_ENV=development hermes dashboard'"
 	@echo ""
-	THEIA_ENV=development cd theia-panel && npx vite --host 127.0.0.1 --port $(DEV_PORT)
+	THEIA_ENV=development cd theia-panel && npx vite --host 0.0.0.0 --port $(DEV_PORT)
 
 dev-panel: ## Start panel dev server only
-	cd theia-panel && npx vite --host 127.0.0.1 --port $(DEV_PORT)
+	cd theia-panel && npx vite --host 0.0.0.0 --port $(DEV_PORT)
 
 dev-link: ## Symlink plugin source into ~/.hermes/plugins for dev
 	@mkdir -p $(HERMES_PLUGINS)
--- a/Makefile
+++ b/Makefile
@@ -69,10 +69,10 @@ dev: dev-link ## Start dev environment (Vite hot-reload + plugin symlink)
 	@echo "    Panel:     http://localhost:$(DEV_PORT)"
 	@echo "    Dashboard: run 'THEIA_ENV=development hermes dashboard'"
 	@echo ""
-	THEIA_ENV=development cd theia-panel && npx vite --host 0.0.0.0 --port $(DEV_PORT)
+	THEIA_ENV=development cd theia-panel && npx vite --host 127.0.0.1 --port $(DEV_PORT)
 
 dev-panel: ## Start panel dev server only
-	cd theia-panel && npx vite --host 0.0.0.0 --port $(DEV_PORT)
+	cd theia-panel && npx vite --host 127.0.0.1 --port $(DEV_PORT)
 
 dev-link: ## Symlink plugin source into ~/.hermes/plugins for dev
 	@mkdir -p $(HERMES_PLUGINS)
--- a/docs/build-pipeline.md
+++ b/docs/build-pipeline.md
@@ -82,7 +82,7 @@ The API validates the dev port before returning it:
 
 | Rule                      | Reason                                    |
 |---------------------------|-------------------------------------------|
-| Must be >= 1024           | Ports < 1024 are system-reserved          |
+| Must be >= 1024           | Ports < 1024 require root privileges on Linux/macOS |
 | Must not be 9119          | Well-known Hermes internal port            |
 
 If validation fails, `dev_panel_url` is set to `null` and
--- a/docs/build-pipeline.md
+++ b/docs/build-pipeline.md
@@ -66,12 +66,14 @@ a custom middleware. `index.html` calls
 
 When `THEIA_ENV=development`, the plugin API's `/config` endpoint returns a
 `dev_panel_url` that the dashboard uses to load the panel iframe. The URL is
-resolved in the following order:
+resolved from the `THEIA_DEV_HOST` environment variable, falling back to
+`localhost`:
 
 1. **`THEIA_DEV_HOST`** env var — explicit override (e.g. `THEIA_DEV_HOST=192.168.1.50`)
-2. **`X-Forwarded-Host`** request header — respects reverse-proxy setups
-3. **`Host`** request header — fallback to the incoming request host
-4. **`localhost`** — ultimate fallback
+2. **`localhost`** — fallback when `THEIA_DEV_HOST` is unset
+
+The API does **not** trust `X-Forwarded-Host` or `Host` request headers for
+host resolution, preventing origin-steering attacks.
 
 Port is controlled by **`THEIA_DEV_PORT`** (default `5173`). Both the host
 and port resolution apply to any environment where `THEIA_ENV=development`.
@@ -83,6 +85,7 @@ The API validates the dev port before returning it:
 | Rule                      | Reason                                    |
 |---------------------------|-------------------------------------------|
 | Must be >= 1024           | Ports < 1024 require root privileges on Linux/macOS |
+| Must be <= 65535          | Valid TCP port range                      |
 | Must not be 9119          | Well-known Hermes internal port            |
 
 If validation fails, `dev_panel_url` is set to `null` and
--- a/docs/build-pipeline.md
+++ b/docs/build-pipeline.md
@@ -62,6 +62,33 @@ During `vite` serve, the dev server reads `theia-graph.json` from
 a custom middleware. `index.html` calls
 `mount(document.getElementById("app"), "/theia-graph.json")`.
 
+### Dev URL resolution
+
+When `THEIA_ENV=development`, the plugin API's `/config` endpoint returns a
+`dev_panel_url` that the dashboard uses to load the panel iframe. The URL is
+resolved in the following order:
+
+1. **`THEIA_DEV_HOST`** env var — explicit override (e.g. `THEIA_DEV_HOST=192.168.1.50`)
+2. **`X-Forwarded-Host`** request header — respects reverse-proxy setups
+3. **`Host`** request header — fallback to the incoming request host
+4. **`localhost`** — ultimate fallback
+
+Port is controlled by **`THEIA_DEV_PORT`** (default `5173`). Both the host
+and port resolution apply to any environment where `THEIA_ENV=development`.
+
+### Port validation
+
+The API validates the dev port before returning it:
+
+| Rule                      | Reason                                    |
+|---------------------------|-------------------------------------------|
+| Must be >= 1024           | Ports < 1024 are system-reserved          |
+| Must not be 9119          | Well-known Hermes internal port            |
+
+If validation fails, `dev_panel_url` is set to `null` and
+`dev_panel_error` contains the reason. This prevents the dashboard from
+loading a misconfigured panel.
+
 ## 4. `theia-panel` library build
 
 ```bash
--- a/examples/dashboard.html
+++ b/examples/dashboard.html
@@ -358,7 +358,7 @@
           </div>
           <iframe
             id="theia-iframe"
-            src="http://localhost:5173/?graph=/graph.json"
+            src="about:blank"
             allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope"
             loading="eager"
           ></iframe>
@@ -367,7 +367,18 @@
     </div>
 
     <script>
+      // ── Configuration ─────────────────────────────────────────────────────
+      // Dev panel URL — override via query param: ?panel=http://host:port
+      const DEFAULT_PANEL_URL = "http://localhost:5173";
+      const panelUrl =
+        new URLSearchParams(window.location.search).get("panel") ||
+        DEFAULT_PANEL_URL;
+      const panelOrigin = new URL(panelUrl).origin;
+
+      // ── Iframe setup ──────────────────────────────────────────────────────
       const iframe = document.getElementById("theia-iframe");
+      iframe.src = panelUrl + "/?graph=/graph.json";
+
       const loading = document.getElementById("iframe-loading");
 
       // Update iframe source display
@@ -394,7 +405,7 @@
 
       // Listen for postMessage from the theia panel
       window.addEventListener("message", (event) => {
-        if (event.origin !== "http://localhost:5173") return;
+        if (event.origin !== panelOrigin) return;
 
         if (event.data?.type === "node-click") {
           console.log("[dashboard] node clicked:", event.data.nodeId);
--- a/examples/dashboard.html
+++ b/examples/dashboard.html
@@ -370,9 +370,17 @@
       // ── Configuration ─────────────────────────────────────────────────────
       // Dev panel URL — override via query param: ?panel=http://host:port
       const DEFAULT_PANEL_URL = "http://localhost:5173";
-      const panelUrl =
+      let panelUrl =
         new URLSearchParams(window.location.search).get("panel") ||
         DEFAULT_PANEL_URL;
+      try {
+        const parsed = new URL(panelUrl);
+        if (parsed.protocol !== "http:" && parsed.protocol !== "https:") {
+          panelUrl = DEFAULT_PANEL_URL;
+        }
+      } catch {
+        panelUrl = DEFAULT_PANEL_URL;
+      }
       const panelOrigin = new URL(panelUrl).origin;
 
       // ── Iframe setup ──────────────────────────────────────────────────────
--- a/examples/dashboard.html
+++ b/examples/dashboard.html
@@ -373,19 +373,22 @@
       let panelUrl =
         new URLSearchParams(window.location.search).get("panel") ||
         DEFAULT_PANEL_URL;
+      let parsed;
       try {
-        const parsed = new URL(panelUrl);
+        parsed = new URL(panelUrl);
         if (parsed.protocol !== "http:" && parsed.protocol !== "https:") {
           panelUrl = DEFAULT_PANEL_URL;
+          parsed = new URL(panelUrl);
         }
       } catch {
         panelUrl = DEFAULT_PANEL_URL;
+        parsed = new URL(panelUrl);
       }
-      const panelOrigin = new URL(panelUrl).origin;
+      const panelOrigin = parsed.origin;
 
       // ── Iframe setup ──────────────────────────────────────────────────────
       const iframe = document.getElementById("theia-iframe");
-      iframe.src = panelUrl + "/?graph=/graph.json";
+      iframe.src = new URL("/?graph=/graph.json", panelUrl).href;
 
       const loading = document.getElementById("iframe-loading");
 
--- a/plugin/api/plugin_api.py
+++ b/plugin/api/plugin_api.py
@@ -35,7 +35,7 @@ log = logging.getLogger("theia-constellation")
 # ---------------------------------------------------------------------------
 
 THEIA_ENV = os.environ.get("THEIA_ENV", "production")
-THEIA_DEV_HOST = os.environ.get("THEIA_DEV_HOST", "")
+_THEIA_DEV_HOST_RAW = os.environ.get("THEIA_DEV_HOST", "")
 _THEIA_DEV_PORT_RAW = os.environ.get("THEIA_DEV_PORT", "5173")
 
 # ---------------------------------------------------------------------------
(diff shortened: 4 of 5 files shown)


plugin/api/plugin_api.py
```diff
@@ -32,24 +35,68 @@
...
     can proxy the iframe to the Vite dev server for hot-reload.
+    The host is derived from the incoming request headers
+    (``x-forwarded-host`` or ``host``), overridable via ``THEIA_DEV_HOST``.
     """
     config: dict = {"env": THEIA_ENV, "version": "0.1.0"}
 
     if THEIA_ENV == "development":
-        config["dev_panel_url"] = f"http://localhost:{THEIA_DEV_PORT}"
+        forwarded = request.headers.get("x-forwarded-host", "")
+        raw_host = request.headers.get("host", "localhost")
```
zebster-cmd: **CRITICAL**: Host header injection: `dev_panel_url` is built from `X-Forwarded-Host` / `Host` headers with no allowlist. An attacker who can reach the API (now easier with the 0.0.0.0 bind) can send `X-Forwarded-Host: evil.tld` and the dashboard will render an iframe pointing at attacker-controlled content, plus accept its postMessages once the dashboard derives `panelOrigin` from the same value. Even when behind a proxy, blindly trusting `X-Forwarded-Host` is a well-known foot-gun.

💡 Validate the resolved host against an allowlist (e.g. `localhost`, `127.0.0.1`, plus any explicit `THEIA_DEV_HOST`), or only consult `X-Forwarded-Host` when an explicit `THEIA_TRUST_PROXY=1` is set. Reject hostnames that don't match `^[A-Za-z0-9.\-]+$` (and bracketed IPv6) before interpolating into the URL.

Makefile
```diff
@@ -69,10 +69,10 @@ dev: dev-link ## Start dev environment (Vite hot-reload + plugin symlink)
 	@echo "    Panel:     http://localhost:$(DEV_PORT)"
 	@echo "    Dashboard: run 'THEIA_ENV=development hermes dashboard'"
 	@echo ""
-	THEIA_ENV=development cd theia-panel && npx vite --host 127.0.0.1 --port $(DEV_PORT)
+	THEIA_ENV=development cd theia-panel && npx vite --host 0.0.0.0 --port $(DEV_PORT)
```
zebster-cmd: **CRITICAL**: Switching the default Vite bind from `127.0.0.1` to `0.0.0.0` exposes the dev server (HMR websocket, source maps, unminified source, /@fs file access) to every interface, including LAN/VPN/Tailscale peers. Combined with the new header-derived `dev_panel_url`, any peer on the network can both reach the dev server and influence the URL the dashboard loads. Defaults should stay loopback-only.

💡 Keep `--host 127.0.0.1` as the default and add an opt-in target (e.g. `dev-lan: HOST?=0.0.0.0`) or document `make dev HOST=0.0.0.0` for the LAN-testing case. Apply the same to `dev-panel` on line 75.

examples/dashboard.html
```diff
@@ -367,7 +367,18 @@ <h2>Integration Notes</h2>
...
+      const DEFAULT_PANEL_URL = "http://localhost:5173";
+      const panelUrl =
+        new URLSearchParams(window.location.search).get("panel") ||
+        DEFAULT_PANEL_URL;
+      const panelOrigin = new URL(panelUrl).origin;
+
+      // ── Iframe setup ──────────────────────────────────────────────────────
       const iframe = document.getElementById("theia-iframe");
+      iframe.src = panelUrl + "/?graph=/graph.json";
+
```
zebster-cmd, line 392: **CRITICAL**: `panelUrl` is taken straight from `?panel=` and assigned to `iframe.src`. `new URL('javascript:alert(1)')` succeeds and yields origin `'null'`, so `?panel=javascript:alert(1)` causes the iframe to execute attacker JS in the dashboard's parent context (DOM-XSS via self-served example). Even if this is only example code, it's the canonical reference users copy.

💡 After parsing, enforce `parsed.protocol === 'http:' || parsed.protocol === 'https:'` and fall back to `DEFAULT_PANEL_URL` otherwise. Wrap the `new URL(panelUrl)` in try/catch to handle malformed inputs without throwing on page load.

plugin/api/plugin_api.py
```diff
@@ -32,24 +35,68 @@
...
+    """Validate and return the port number.
+
+    Raises ``ValueError`` if the port is out of range or blocked.
+    """
+    try:
+        p = int(port)
+    except (ValueError, TypeError):
+        raise ValueError(f"Invalid port: {port!r}")
+
+    if p < 1024:
```
zebster-cmd, line 61: **WARNING**: Port validation has no upper bound — `THEIA_DEV_PORT=99999` passes the `>= 1024` and not-blocked checks and produces an invalid URL. Also `int('5173 ')` succeeds but `int('5173\n')` raises; the error path is fine but bounds are incomplete.

💡 Add `if p > 65535: raise ValueError(...)` next to the lower-bound check, and use `raise ValueError(...) from None` (or `from e`) inside the `except` block to keep tracebacks clean.

plugin/api/plugin_api.py
zebster-cmd: **WARNING**: `(forwarded or raw_host).split(':')[0]` mishandles IPv6 hosts like `[::1]:5173` — it returns `'['`, producing a broken URL `http://[:port`. Unlikely in dev but trivially triggered if a developer hits the API over IPv6 loopback.

💡 Use `urllib.parse` or detect the bracketed form: if the value starts with `[`, split on `']'` and re-add the bracket; otherwise rsplit on `':'` once. Better yet, parse with `urllib.parse.urlsplit('//'+value).hostname`.

examples/dashboard.html
```diff
@@ -367,7 +367,18 @@ <h2>Integration Notes</h2>
...
 
     <script>
+      // ── Configuration ─────────────────────────────────────────────────────
+      // Dev panel URL — override via query param: ?panel=http://host:port
+      const DEFAULT_PANEL_URL = "http://localhost:5173";
+      const panelUrl =
+        new URLSearchParams(window.location.search).get("panel") ||
+        DEFAULT_PANEL_URL;
+      const panelOrigin = new URL(panelUrl).origin;
+
```
zebster-cmd, line 388: **WARNING**: `new URL(panelUrl).origin` runs unconditionally during script init — a malformed `?panel=` query crashes the script before the iframe is wired up, leaving the loading spinner stuck and silent.

💡 Wrap URL parsing in try/catch and fall back to `DEFAULT_PANEL_URL` (also fixes the javascript:-scheme issue above).

plugin/api/plugin_api.py
```diff
@@ -32,24 +35,68 @@
...
 
 THEIA_ENV = os.environ.get("THEIA_ENV", "production")
+THEIA_DEV_HOST = os.environ.get("THEIA_DEV_HOST", "")
 THEIA_DEV_PORT = os.environ.get("THEIA_DEV_PORT", "5173")
 
+# ---------------------------------------------------------------------------
+# Helpers
+# ---------------------------------------------------------------------------
+
+_BLOCKED_PORTS: set[int] = {9119}
```
zebster-cmd: **SUGGESTION**: `_BLOCKED_PORTS: set[int] = {9119}` is mutable module-level state — anything imported into the same process can mutate it. Prefer `frozenset` for an intentionally read-only constant.

💡 `_BLOCKED_PORTS: frozenset[int] = frozenset({9119})`

plugin/api/plugin_api.py
```diff
@@ -32,24 +35,68 @@
...
 
     if THEIA_ENV == "development":
-        config["dev_panel_url"] = f"http://localhost:{THEIA_DEV_PORT}"
+        forwarded = request.headers.get("x-forwarded-host", "")
+        raw_host = request.headers.get("host", "localhost")
+        host = (forwarded or raw_host).split(":")[0]
+
+        try:
+            config["dev_panel_url"] = _resolve_dev_url(host)
+        except ValueError as e:
```
zebster-cmd, line 145: **SUGGESTION**: Validation failure is logged at WARNING on every `/config` hit. If a misconfigured deploy stays misconfigured, this will spam logs. Consider logging once at startup (validate `THEIA_DEV_PORT` when the module loads) and just returning the cached error from the handler.

💡 Compute `_DEV_PORT_VALIDATED` / `_DEV_PORT_ERROR` at import time when `THEIA_ENV == 'development'`; log once; the handler just reads the cached values.

docs/build-pipeline.md
```diff
@@ -62,6 +62,33 @@ During `vite` serve, the dev server reads `theia-graph.json` from
...
+
+Port is controlled by **`THEIA_DEV_PORT`** (default `5173`). Both the host
+and port resolution apply to any environment where `THEIA_ENV=development`.
+
+### Port validation
+
+The API validates the dev port before returning it:
+
+| Rule                      | Reason                                    |
+|---------------------------|-------------------------------------------|
```
zebster-cmd, line 86: **SUGGESTION**: Doc says ports < 1024 are 'system-reserved'. Technically those are IANA system/well-known ports — the practical reason to block them is that binding requires root on Linux/macOS. Minor wording nit but worth fixing for accuracy.

💡 Reword to: 'Ports < 1024 require elevated privileges to bind on Unix systems.'

zebster-cmd (changes requested):
## Code Review

The change introduces a useful dev-host override mechanism, but it ships two notable security regressions: the Vite dev server is now bound to 0.0.0.0 (LAN-exposed by default), and `/config` reflects untrusted `X-Forwarded-Host` / `Host` headers straight into `dev_panel_url` with no allowlist — letting an attacker who can reach the API steer the dashboard's iframe at an arbitrary origin. There are also a handful of input-validation gaps (no upper port bound, IPv6 host parsing, javascript: URLs accepted by the example dashboard) worth tightening before merge.

zebster-cmd:
## Code Review Summary

**Verdict:** REQUEST_CHANGES

The change introduces a useful dev-host override mechanism, but it ships two notable security regressions: the Vite dev server is now bound to 0.0.0.0 (LAN-exposed by default), and `/config` reflects untrusted `X-Forwarded-Host` / `Host` headers straight into `dev_panel_url` with no allowlist — letting an attacker who can reach the API steer the dashboard's iframe at an arbitrary origin. There are also a handful of input-validation gaps (no upper port bound, IPv6 host parsing, javascript: URLs accepted by the example dashboard) worth tightening before merge.

### Findings
- 🔴 **plugin/api/plugin_api.py:91** — Host header injection: `dev_panel_url` is built from `X-Forwarded-Host` / `Host` headers with no allowlist. An attacker who can reach the API (now easier with the 0.0.0.0 bind) can send `X-Forwarded-Host: evil.tld` and the dashboard will render an iframe pointing at attacker-controlled content, plus accept its postMessages once the dashboard derives `panelOrigin` from the same value. Even when behind a proxy, blindly trusting `X-Forwarded-Host` is a well-known foot-gun.
- 🔴 **Makefile:72** — Switching the default Vite bind from `127.0.0.1` to `0.0.0.0` exposes the dev server (HMR websocket, source maps, unminified source, /@fs file access) to every interface, including LAN/VPN/Tailscale peers. Combined with the new header-derived `dev_panel_url`, any peer on the network can both reach the dev server and influence the URL the dashboard loads. Defaults should stay loopback-only.
- 🔴 **examples/dashboard.html:381** — `panelUrl` is taken straight from `?panel=` and assigned to `iframe.src`. `new URL('javascript:alert(1)')` succeeds and yields origin `'null'`, so `?panel=javascript:alert(1)` causes the iframe to execute attacker JS in the dashboard's parent context (DOM-XSS via self-served example). Even if this is only example code, it's the canonical reference users copy.
- ⚠️ **plugin/api/plugin_api.py:58** — Port validation has no upper bound — `THEIA_DEV_PORT=99999` passes the `>= 1024` and not-blocked checks and produces an invalid URL. Also `int('5173 ')` succeeds but `int('5173\n')` raises; the error path is fine but bounds are incomplete.
- ⚠️ **plugin/api/plugin_api.py:91** — `(forwarded or raw_host).split(':')[0]` mishandles IPv6 hosts like `[::1]:5173` — it returns `'['`, producing a broken URL `http://[:port`. Unlikely in dev but trivially triggered if a developer hits the API over IPv6 loopback.
- ⚠️ **examples/dashboard.html:377** — `new URL(panelUrl).origin` runs unconditionally during script init — a malformed `?panel=` query crashes the script before the iframe is wired up, leaving the loading spinner stuck and silent.
- 💡 **plugin/api/plugin_api.py:45** — `_BLOCKED_PORTS: set[int] = {9119}` is mutable module-level state — anything imported into the same process can mutate it. Prefer `frozenset` for an intentionally read-only constant.
- 💡 **plugin/api/plugin_api.py:96** — Validation failure is logged at WARNING on every `/config` hit. If a misconfigured deploy stays misconfigured, this will spam logs. Consider logging once at startup (validate `THEIA_DEV_PORT` when the module loads) and just returning the cached error from the handler.
- 💡 **docs/build-pipeline.md:84** — Doc says ports < 1024 are 'system-reserved'. Technically those are IANA system/well-known ports — the practical reason to block them is that binding requires root on Linux/macOS. Minor wording nit but worth fixing for accuracy.

zebster-cmd (commented):
# Follow-up review — PR #48 (commit `fe20666`)

### Summary
`fe20666` walks back the header-derived `dev_panel_url` (now defaults to `localhost`, overridable only via `THEIA_DEV_HOST`), reverts the Vite bind to `127.0.0.1`, adds a `> 65535` port check + IPv6-aware host parsing, sanitizes the dashboard `?panel=` query, and moves port validation to import-time so `/config` no longer log-spams.

### Verification of previous review items

1. **`0.0.0.0` default reverted?** — **Addressed.** Both `dev` and `dev-panel` targets now pass `--host 127.0.0.1` (`Makefile:72`, `Makefile:75`).

2. **Host allowlist added?** — **Addressed in spirit, mis-described.** Untrusted headers are no longer read; `host = THEIA_DEV_HOST or "localhost"` (`plugin/api/plugin_api.py:120-121`). This closes the open-redirect/iframe-hijack vector. However, the commit message ("add host allowlist") and PR body ("`make dev` now binds to `0.0.0.0` by default") are now stale — there is *no* allowlist; there is a single env override. Title/description should be updated.

3. **Port upper bound?** — **Addressed.** `plugin/api/plugin_api.py:63-64` rejects `p > 65535` with a clear message; docstring at `plugin/api/plugin_api.py:16-17` reflects the new rule.

4. **IPv6 parsing fix?** — **Addressed.** `_extract_host` (`plugin/api/plugin_api.py:78-84`) handles `[::1]:5173` by splitting on `]` and stripping `[`, and `_format_host_for_url` (`plugin/api/plugin_api.py:87-91`) re-brackets bare IPv6 literals. The previous `(forwarded or raw_host).split(':')[0]` foot-gun is gone (`plugin/api/plugin_api.py:121`).

5. **Dashboard URL param sanitization?** — **Addressed.** `examples/dashboard.html:373-383` parses `?panel=` inside `try/catch` and forces `DEFAULT_PANEL_URL` unless `protocol` is `http:`/`https:`. This blocks `javascript:` (DOM-XSS) **and** the malformed-URL crash that left the spinner stuck.

6. **`_BLOCKED_PORTS` → frozenset?** — **Addressed.** `plugin/api/plugin_api.py:45`.

7. **Per-request log spam on bad port?** — **Addressed.** Validation now runs once at import (`plugin/api/plugin_api.py:71-75`); `_resolve_dev_url` re-raises the cached error (`plugin/api/plugin_api.py:96-97`), and the per-request `log.warning` is removed (`plugin/api/plugin_api.py:125-127`).

8. **Doc wording on `<1024`?** — **Addressed.** `docs/build-pipeline.md:85`.

### New issues introduced by fe20666

- **`_extract_host` returns empty string for un-bracketed IPv6** (`plugin/api/plugin_api.py:78-84`). If a user sets `THEIA_DEV_HOST=::1` (no brackets) the `host.split(":")[0]` branch returns `''`, producing `http://[]:5173` after `_format_host_for_url`. Mis-config only — but trivially avoidable.
- **`THEIA_DEV_HOST` is unvalidated** (`plugin/api/plugin_api.py:38`, `plugin/api/plugin_api.py:121`). A value like `panel.example.com:5173` is silently truncated by `_extract_host`; whitespace / empty / non-hostname garbage is accepted. Consider validating at import next to the port check.
- **PR body drift** (not code): the PR description still claims "`make dev` now binds to `0.0.0.0` by default" — directly contradicts `Makefile:72`. Will mislead anyone reading just the PR summary.
- **Redundant `new URL(panelUrl)` parse** at `examples/dashboard.html:384` (already parsed at `examples/dashboard.html:377`). Cosmetic; reuse `parsed.origin`.

### Residual concerns

- **No tests** added for `_validate_port` boundaries (1023 / 1024 / 65535 / 65536 / 9119), `_extract_host` (IPv4, IPv4:port, `[::1]`, `[::1]:p`, bare `::1`), or the dashboard sanitizer. All net-new logic; easy unit-test surface.
- **`iframe.src = panelUrl + "/?graph=/graph.json"`** (`examples/dashboard.html:388`) still does naive string concat — if a future default ends with `/` you get `//?graph=`. Not flagged before, still worth a `new URL(...)` build.
- **CSRF/CORS** on `/config` was out of scope for the prior review and remains so.

### Suggestions

1. Tighten `_extract_host` for the un-bracketed IPv6 case:
   ```python
   if host.count(":") > 1 and not host.startswith("["):
       return host  # bare IPv6, no port
   return host.split(":")[0]
   ```
2. Validate `THEIA_DEV_HOST` at import next to `_DEV_PORT` and cache the error symmetrically.
3. Update PR body/commit message — drop the "host allowlist" / "0.0.0.0 default" lines so the merge commit history matches reality.
4. Add pytest cases for `_validate_port` (boundaries) and `_extract_host` / `_format_host_for_url` round-trips.

### Verdict

**Approve with nits.** Every blocker from the prior review is fixed in code; remaining items are a stale PR description, two minor edge cases, and missing tests — none of which justify another round of `Request changes`.

PR #53, opened 2026-04-25
fix: wrap pitch rotation instead of clamping to [90, -90]

### LINKED ISSUE #49: bug: pitch rotation clamped [90, -90] instead of wrapping
## Bug Report

Pitch rotation is clamped to the range [90, -90] degrees instead of wrapping around. This causes a sudden snap when crossing the zenith/nadir rather than smooth continuous rotation.

### Expected behavior

Pitch should wrap around (e.g., go from 89 → 91 or -89 → -91) without snapping back to the opposite limit.

### Actual behavior

Pitch gets clamped at ±90°, causing a hard stop / sudden flip.

arm64be:
Closes #49

### DIFF
--- a/theia-core/theia_core/detect/cron.py
+++ b/theia-core/theia_core/detect/cron.py
@@ -21,7 +21,7 @@ def detect_cron_chain(sessions: list[Session]) -> list[Edge]:
     edges: list[Edge] = []
     for job_id, group in by_job.items():
         group.sort(key=lambda s: s.started_at)
-        for prev_sess, next_sess in zip(group, group[1:], strict=True):
+        for prev_sess, next_sess in zip(group, group[1:]):
             hours = (next_sess.started_at - prev_sess.started_at).total_seconds() / 3600
             weight = max(0.1, min(1.0, 1.0 / (1.0 + hours * 0.1)))
             edges.append(
--- a/theia-core/theia_core/detect/cron.py
+++ b/theia-core/theia_core/detect/cron.py
@@ -1,0 +1,39 @@
+from __future__ import annotations
+
+from collections import defaultdict
+
+from theia_core.detect import Edge
+from theia_core.ingest import Session
+
+
+def detect_cron_chain(sessions: list[Session]) -> list[Edge]:
+    """Create edges between consecutive runs of the same cron job.
+
+    Sessions spawned by the same cron job are grouped by ``cron_job_id``,
+    sorted chronologically, and an edge is created between each consecutive
+    pair.  This lets users visually trace recurring task lineages.
+    """
+    by_job: dict[str, list[Session]] = defaultdict(list)
+    for sess in sessions:
+        if sess.cron_job_id:
+            by_job[sess.cron_job_id].append(sess)
+
+    edges: list[Edge] = []
+    for job_id, group in by_job.items():
+        group.sort(key=lambda s: s.started_at)
+        for prev_sess, next_sess in zip(group, group[1:], strict=True):
+            hours = (next_sess.started_at - prev_sess.started_at).total_seconds() / 3600
+            weight = max(0.1, min(1.0, 1.0 / (1.0 + hours * 0.1)))
+            edges.append(
+                Edge(
+                    source=prev_sess.id,
+                    target=next_sess.id,
+                    kind="cron-chain",
+                    weight=weight,
+                    evidence={
+                        "cron_job_id": job_id,
+                        "interval_hours": round(hours, 2),
+                    },
+                )
+            )
+    return edges
--- a/theia-core/theia_core/detect/cron.py
+++ b/theia-core/theia_core/detect/cron.py
@@ -21,7 +21,7 @@ def detect_cron_chain(sessions: list[Session]) -> list[Edge]:
     edges: list[Edge] = []
     for job_id, group in by_job.items():
         group.sort(key=lambda s: s.started_at)
-        for prev_sess, next_sess in zip(group, group[1:]):
+        for prev_sess, next_sess in zip(group, group[1:], strict=False):
             hours = (next_sess.started_at - prev_sess.started_at).total_seconds() / 3600
             weight = max(0.1, min(1.0, 1.0 / (1.0 + hours * 0.1)))
             edges.append(
--- a/theia-core/theia_core/ingest.py
+++ b/theia-core/theia_core/ingest.py
@@ -44,6 +44,7 @@ class Session:
     preview: str = ""
     raw: dict[str, Any] = field(default_factory=dict)
     parent_id: str | None = None
+    cron_job_id: str | None = None
 
 
 def _parse_iso(s: str) -> datetime:
@@ -178,6 +179,20 @@ def _build_preview(messages: list[dict[str, Any]]) -> str:
     return ""
 
 
+def _extract_cron_job_id(session_id: str) -> str | None:
+    """Extract cron job ID from a cron-spawned session ID.
+
+    Cron session IDs follow the pattern ``cron_<job_id>_<timestamp>``,
+    e.g. ``cron_8974a4d0b3cc_20260330_171037``.
+    """
+    if not session_id.startswith("cron_"):
+        return None
+    parts = session_id.split("_", 2)
+    if len(parts) < 3:
+        return None
+    return parts[1]
+
+
 def load_sessions(db_path: Path, include_children: bool = True) -> list[Session]:
     """Load sessions from a Hermes SQLite database.
 
@@ -254,6 +269,8 @@ def load_sessions(db_path: Path, include_children: bool = True) -> list[Session]
             ended_at = _datetime_from_timestamp(row["ended_at"])
             duration_sec = max(0.0, (ended_at - started_at).total_seconds())
 
+            cron_job_id = _extract_cron_job_id(session_id)
+
             sessions.append(
                 Session(
                     id=session_id,
@@ -268,6 +285,7 @@ def load_sessions(db_path: Path, include_children: bool = True) -> list[Session]
                     preview=preview,
                     raw={"db_row": dict(row), "messages": messages},
                     parent_id=row["parent_session_id"],
+                    cron_job_id=cron_job_id,
                 )
             )
 
--- a/theia-panel/src/index.ts
+++ b/theia-panel/src/index.ts
@@ -132,7 +132,7 @@ export async function mount(
     modelFilter?: string | null,
   ): Set<string> {
     const kindVisible = new Set<string>();
-    if (enabledKinds.has("subagent")) {
+    if (enabledKinds.has("subagent") || enabledKinds.has("cron-chain")) {
       for (const node of graph.nodes) {
         kindVisible.add(node.id);
       }
--- a/theia-panel/src/index.ts
+++ b/theia-panel/src/index.ts
@@ -132,7 +132,7 @@ export async function mount(
     modelFilter?: string | null,
   ): Set<string> {
     const kindVisible = new Set<string>();
-    if (enabledKinds.has("subagent") || enabledKinds.has("cron-chain")) {
+    if (enabledKinds.has("subagent")) {
       for (const node of graph.nodes) {
         kindVisible.add(node.id);
       }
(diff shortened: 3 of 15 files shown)


theia-core/theia_core/detect/cron.py
```diff
@@ -0,0 +1,39 @@
...
+    """
+    by_job: dict[str, list[Session]] = defaultdict(list)
+    for sess in sessions:
+        if sess.cron_job_id:
+            by_job[sess.cron_job_id].append(sess)
+
+    edges: list[Edge] = []
+    for job_id, group in by_job.items():
+        group.sort(key=lambda s: s.started_at)
+        for prev_sess, next_sess in zip(group, group[1:], strict=True):
```
zebster-cmd: **CRITICAL**: `zip(group, group[1:], strict=True)` will always raise ValueError when group is non-empty. The two iterables differ in length by 1 by construction, and strict=True requires equal lengths. Since by_job only contains non-empty groups, every call to detect_cron_chain on real data will crash before producing edges.

💡 Use `itertools.pairwise(group)` (Python 3.10+, repo requires 3.11+) or drop strict=True: `for prev_sess, next_sess in pairwise(group):`. pairwise is also clearer about intent.

theia-panel/src/index.ts
```diff
@@ -132,7 +132,7 @@ export async function mount(
     modelFilter?: string | null,
   ): Set<string> {
     const kindVisible = new Set<string>();
-    if (enabledKinds.has("subagent")) {
+    if (enabledKinds.has("subagent") || enabledKinds.has("cron-chain")) {
```
zebster-cmd: **WARNING**: Adding `|| enabledKinds.has("cron-chain")` makes enabling cron-chain show ALL nodes (including subagent children that should be hidden when subagent is off). The original `if (enabledKinds.has("subagent"))` branch exists specifically so subagent edges drag in their endpoints; cron-chain has no such asymmetry — its endpoints are root sessions and the existing else-branch already handles them correctly via hasNonSubagentConnection.

💡 Revert this line to `if (enabledKinds.has("subagent")) {`. cron-chain edges will be picked up by the `for (const edge of graph.edges)` loop in the else branch like every other non-subagent kind.

theia-panel/src/ui/SidePanel.ts
```diff
@@ -280,6 +280,13 @@ function renderEdge(
     detail = `<div style="${detailAttrStyle(theme)}">
       ${label}: <span style="color:#${theme.accent}">${escape(String(displayId ?? "?"))}</span>
     </div>`;
+  } else if (e.kind === "cron-chain") {
+    const jobId = ev.cron_job_id;
+    const intervalHours = ev.interval_hours;
+    detail = `<div style="${detailAttrStyle(theme)}">
+      cron job: <span style="color:#${theme.accent}">${escape(String(jobId ?? "?"))}</span>
+      ${intervalHours !== undefined ? `\u00b7 interval ${Number(intervalHours).toFixed(1)}h` : ""}
```
zebster-cmd, line 288: **WARNING**: `ev.interval_hours` is `unknown` (evidence is `Record<string, unknown>`). The `!== undefined` guard passes for null and non-numeric values, and `Number(non-numeric).toFixed(1)` yields the string "NaN". Today the producer in cron.py always sends a number, but the panel should defensively validate at the boundary like the rest of the unknown-evidence handling.

💡 Narrow before formatting: `const n = typeof intervalHours === "number" ? intervalHours : Number(intervalHours); ... ${Number.isFinite(n) ? `· interval ${n.toFixed(1)}h` : ""}`.

theia-panel/src/scene/Scene.ts
```diff
@@ -85,7 +85,14 @@ export function createScene(container: HTMLElement): SceneContext {
...
       theta -= dxPixel * rotateSpeed;
       phi -= dyPixel * rotateSpeed;
-      phi = Math.max(0.01, Math.min(Math.PI - 0.01, phi));
+      if (phi < 0) {
+        phi = -phi;
+        theta += Math.PI;
+      } else if (phi > Math.PI) {
+        phi = 2 * Math.PI - phi;
+        theta += Math.PI;
+      }
```
zebster-cmd, line 94: **SUGGESTION**: The pole-reflection change (and the simultaneous tightening of the clamp from 0.01 to 0.001) is unrelated to cron-chain detection and should ideally land in its own PR with a separate justification — bundling it makes the feature diff harder to review and bisect.

💡 Split the rotate() pole-handling change into a separate commit/PR. Also note: a single drag delta that overshoots by more than π (e.g. fast trackpad fling) reflects to a value still outside [0, π], the elif isn't taken, and it gets clamped to the boundary. Consider `phi = ((phi % (2*Math.PI)) + 2*Math.PI) % (2*Math.PI)` then reflect, or normalize iteratively.

theia-core/theia_core/detect/cron.py
```diff
@@ -0,0 +1,39 @@
+from __future__ import annotations
+
+from collections import defaultdict
+
+from theia_core.detect import Edge
+from theia_core.ingest import Session
+
+
+def detect_cron_chain(sessions: list[Session]) -> list[Edge]:
```
zebster-cmd, line 9: **SUGGESTION**: No tests accompany this new detector even though the existing detect/* modules are amenable to pure-function testing. Given the strict=True bug above slipped through, a small unit test (e.g. two sessions with same cron_job_id produce one edge; one session produces zero edges; two jobs don't cross-link) would have caught this and is cheap to write.

💡 Add a test under theia-core/tests/ exercising at least: zero, one, two, and three sessions per cron_job_id, plus the timestamp ordering and the weight formula bounds (0.1 ≤ w ≤ 1.0).

theia-core/theia_core/ingest.py
```diff
@@ -178,6 +179,20 @@ def _build_preview(messages: list[dict[str, Any]]) -> str:
...
+    """Extract cron job ID from a cron-spawned session ID.
+
+    Cron session IDs follow the pattern ``cron_<job_id>_<timestamp>``,
+    e.g. ``cron_8974a4d0b3cc_20260330_171037``.
+    """
+    if not session_id.startswith("cron_"):
+        return None
+    parts = session_id.split("_", 2)
+    if len(parts) < 3:
+        return None
```
zebster-cmd, line 192: **SUGGESTION**: `_extract_cron_job_id` does not validate the timestamp segment, so any session id starting with `cron_<x>_<anything>` parses as a cron job. If a non-cron session ID happens to share that prefix, it'll get bucketed and produce spurious cron-chain edges. Worth tightening if the ID scheme isn't strictly reserved.

💡 Either document `cron_` as a reserved prefix in the ingest contract, or validate the third segment looks like a timestamp (e.g. `re.match(r"\d{8}_\d{6}", parts[2])`) before returning the job id.

zebster-cmd (changes requested):
## Code Review

The cron-chain feature is wired through cleanly across schema, types, palette, physics, filters, and side panel, but `detect_cron_chain` contains a critical bug: `zip(group, group[1:], strict=True)` raises `ValueError` whenever a cron job has any sessions, since the two iterables differ in length by 1. The visibility-logic change in index.ts also unintentionally bypasses subagent filtering whenever cron-chain is enabled, and a tangential camera-pole reflection change is bundled in. Once the zip bug is fixed (and ideally backed by a test), the rest is approachable.

zebster-cmd:
## Code Review Summary

**Verdict:** REQUEST_CHANGES

The cron-chain feature is wired through cleanly across schema, types, palette, physics, filters, and side panel, but `detect_cron_chain` contains a critical bug: `zip(group, group[1:], strict=True)` raises `ValueError` whenever a cron job has any sessions, since the two iterables differ in length by 1. The visibility-logic change in index.ts also unintentionally bypasses subagent filtering whenever cron-chain is enabled, and a tangential camera-pole reflection change is bundled in. Once the zip bug is fixed (and ideally backed by a test), the rest is approachable.

### Findings
- 🔴 **theia-core/theia_core/detect/cron.py:24** — `zip(group, group[1:], strict=True)` will always raise ValueError when group is non-empty. The two iterables differ in length by 1 by construction, and strict=True requires equal lengths. Since by_job only contains non-empty groups, every call to detect_cron_chain on real data will crash before producing edges.
- ⚠️ **theia-panel/src/index.ts:135** — Adding `|| enabledKinds.has("cron-chain")` makes enabling cron-chain show ALL nodes (including subagent children that should be hidden when subagent is off). The original `if (enabledKinds.has("subagent"))` branch exists specifically so subagent edges drag in their endpoints; cron-chain has no such asymmetry — its endpoints are root sessions and the existing else-branch already handles them correctly via hasNonSubagentConnection.
- ⚠️ **theia-panel/src/ui/SidePanel.ts:288** — `ev.interval_hours` is `unknown` (evidence is `Record<string, unknown>`). The `!== undefined` guard passes for null and non-numeric values, and `Number(non-numeric).toFixed(1)` yields the string "NaN". Today the producer in cron.py always sends a number, but the panel should defensively validate at the boundary like the rest of the unknown-evidence handling.
- 💡 **theia-panel/src/scene/Scene.ts:94** — The pole-reflection change (and the simultaneous tightening of the clamp from 0.01 to 0.001) is unrelated to cron-chain detection and should ideally land in its own PR with a separate justification — bundling it makes the feature diff harder to review and bisect.
- 💡 **theia-core/theia_core/detect/cron.py:9** — No tests accompany this new detector even though the existing detect/* modules are amenable to pure-function testing. Given the strict=True bug above slipped through, a small unit test (e.g. two sessions with same cron_job_id produce one edge; one session produces zero edges; two jobs don't cross-link) would have caught this and is cheap to write.
- 💡 **theia-core/theia_core/ingest.py:192** — `_extract_cron_job_id` does not validate the timestamp segment, so any session id starting with `cron_<x>_<anything>` parses as a cron job. If a non-cron session ID happens to share that prefix, it'll get bucketed and produce spurious cron-chain edges. Worth tightening if the ID scheme isn't strictly reserved.

zebster-cmd:
Closed as effectively merged via PR #54 — partially literal, partially via parallel commits — with the unique surviving content salvaged into #65.

**Literal merge via #54:** The pitch-rotation fix commit `6ef4e49` from this branch was carried into `main` as part of PR #54's merge commit `0f74cb7`. This was unintentional cross-contamination — `6ef4e49` should have shipped through this PR (#53), not via #54 — but the code itself is the same.

**Content-equivalent on main via parallel commits:** The intermediate review-fix commits on this branch (`9b5b5ac`, `29bd023`, `745eb86`, `42226e5`) are not literally on `main`, but every code change they propose has an equivalent or improved counterpart on `main`, delivered by independent commits authored during PR #54's review cycle (`f122d66`, `e91cbc0`, `0cfad35`, etc. — same pattern as documented in #47's close comment).

**Salvaged into #65:** The only unique surviving content on this branch was `theia-core/tests/test_cron.py` (89 lines, 6 unit tests for `detect_cron_chain()`). Those tests have been salvaged into a focused PR at #65, re-authored as `zebster-cmd <zeb@ipster.nl>` per the project's commit-author convention. All 6 tests pass against current `main`.

**Why salvage instead of rebase-in-place:** Rebasing this branch onto `main` would have force-pushed past a now-misleading PR title ("fix: wrap pitch rotation") and marked most of the original inline review comments as "Outdated" since the lines they anchored to are gone. Salvaging keeps this PR's review thread intact as the canonical record of the pitch-rotation discussion (which actually shipped via #54), and gives the cron-chain tests a clean review surface in #65.

GitHub will mark this as `Closed` rather than `Merged` because the branch tip SHA (`42226e5`) isn't reachable from `main` — but the work itself shipped (pitch fix via #54, tests via #65 once that lands).

**Process retrospective.** This branch and #54's branch shared a commit ancestry (`4690069 → 6ef4e49 → 9dfcc82 → dd31ff2`) instead of branching independently from `main`. Going forward, any PR review on this repo should start with `git log <base>..<branch>` checked against the PR's stated scope, and feature branches should be created off `main` directly to prevent this. See #47's close comment for the parallel cron-chain instance.

zebster-cmd:
PR #65 has been reviewed. See review comments for details.

PR #55, opened 2026-04-25
docs: add curl-pipe-shell one-liner as primary install method in README

### LINKED ISSUE #51: README should suggest curl-pipe-shell one-liner for install
## Description

Now that `install.sh` pulls the git repo (instead of a tarball), the README's primary install instruction should be a one-liner:

```sh
curl -fsSL https://raw.githubusercontent.com/arm64be/theia/main/install.sh | sh
```

This depends on the install.sh migration (tarball → git clone).

arm64be:
Closes #51

Depends on #50 (install.sh must support git clone for the one-liner to work).

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,16 @@ Visualize Hermes agent sessions as a semantic constellation.
 - `theia-panel/` — TypeScript panel: `theia-graph.json` → three.js constellation
 - `schemas/graph.schema.json` — contract between the two halves
 
+## Install
+
+The fastest way to install theia:
+
+```bash
+curl -fsSL https://raw.githubusercontent.com/arm64be/theia/main/install.sh | sh
+```
+
+See [`docs/quick-start.md`](docs/quick-start.md) for alternative install methods, prerequisites, and configuration.
+
 ## Documentation
 
 | Document | Description |


theia-core/theia_core/detect/cron.py
```diff
@@ -0,0 +1,39 @@
...
+    """
+    by_job: dict[str, list[Session]] = defaultdict(list)
+    for sess in sessions:
+        if sess.cron_job_id:
+            by_job[sess.cron_job_id].append(sess)
+
+    edges: list[Edge] = []
+    for job_id, group in by_job.items():
+        group.sort(key=lambda s: s.started_at)
+        for prev_sess, next_sess in zip(group, group[1:], strict=True):
```
zebster-cmd: **CRITICAL**: `strict=True` will raise ValueError every time this loop runs with a non-trivial group. `group` and `group[1:]` always differ in length by exactly one, so as soon as the shorter iterable is exhausted, strict-zip will detect the mismatch and raise. This means cron-chain edges will never actually be emitted (and worse, the build pipeline will crash whenever any cron job has ≥2 sessions).

💡 Drop `strict=True` (default zip stops at the shortest iterable, which is the desired behavior here), or use `itertools.pairwise(group)` which is the idiomatic way to walk consecutive pairs:

```python
from itertools import pairwise
...
for prev_sess, next_sess in pairwise(group):
```
Also please add a unit test covering a job with 2+ sessions — this would have caught it immediately.

install.sh
```diff
@@ -40,16 +40,14 @@ REPO_TAG="main"  # pin to a release tag or commit SHA for production
...
     TMPDIR="$(mktemp -d)"
-    ARCHIVE_URL="${REPO_URL}/archive/${REPO_TAG}.tar.gz"
 
     echo ""
-    echo "  Downloading Theia from ${ARCHIVE_URL} ..."
+    echo "  Cloning Theia from ${REPO_URL} ..."
     echo ""
 
-    curl -fsSL "$ARCHIVE_URL" | tar xz -C "$TMPDIR"
+    git clone --depth 1 --branch "$REPO_TAG" "$REPO_URL" "$TMPDIR"
```
zebster-cmd, line 54: **WARNING**: Switching from tarball-download to `git clone --depth 1 --branch "$REPO_TAG"` regresses commit-SHA support. The inline comment on REPO_TAG still says `pin to a release tag or commit SHA for production`, but `git clone --branch` only accepts branch and tag refs — pinning REPO_TAG to a commit SHA will now fail with `Remote branch <sha> not found`. The previous tarball URL `archive/<sha>.tar.gz` worked for any ref. This also adds a hard dependency on `git` being installed on the user's machine for the curl-piped install path, where previously only curl+tar were needed.

💡 Either (a) update the comment to say tag/branch only, or (b) keep SHA support by doing `git clone --depth 1 "$REPO_URL" "$TMPDIR" && git -C "$TMPDIR" fetch --depth 1 origin "$REPO_TAG" && git -C "$TMPDIR" checkout FETCH_HEAD`, or revert to the tarball approach which handles all ref types and avoids the git dependency.

theia-panel/src/ui/SidePanel.ts
```diff
@@ -280,6 +280,13 @@ function renderEdge(
     detail = `<div style="${detailAttrStyle(theme)}">
       ${label}: <span style="color:#${theme.accent}">${escape(String(displayId ?? "?"))}</span>
     </div>`;
+  } else if (e.kind === "cron-chain") {
+    const jobId = ev.cron_job_id;
+    const intervalHours = ev.interval_hours;
+    detail = `<div style="${detailAttrStyle(theme)}">
+      cron job: <span style="color:#${theme.accent}">${escape(String(jobId ?? "?"))}</span>
+      ${intervalHours !== undefined ? `\u00b7 interval ${Number(intervalHours).toFixed(1)}h` : ""}
```
zebster-cmd: **SUGGESTION**: `ev.interval_hours` is typed `unknown` (evidence is `Record<string, unknown>`), so `Number(intervalHours).toFixed(1)` can render the string "NaN" if the field is ever missing or non-numeric. The `!== undefined` check doesn't catch null or non-numeric strings.

💡 Validate it's a finite number before formatting:

```ts
const hoursNum = typeof intervalHours === "number" ? intervalHours : Number(intervalHours);
const intervalSuffix = Number.isFinite(hoursNum) ? ` · interval ${hoursNum.toFixed(1)}h` : "";
```

theia-core/theia_core/detect/cron.py
```diff
@@ -0,0 +1,39 @@
...
+            hours = (next_sess.started_at - prev_sess.started_at).total_seconds() / 3600
+            weight = max(0.1, min(1.0, 1.0 / (1.0 + hours * 0.1)))
```
zebster-cmd: **SUGGESTION**: The weight formula `1 / (1 + hours * 0.1)` is undocumented. A future reader (or you, in 6 months) won't know whether the 0.1 decay constant is intentional or a magic number — and at hours=0 (clock-skewed sessions or simultaneous runs) weight is 1.0, which equals the strongest semantic edges in the system. Consider whether that's intended.

💡 Add a short comment explaining the curve ("weight ≈ 0.5 at 10h interval, clamped to [0.1, 1.0]") or extract the decay constant to a named module-level value so it's adjustable without re-deriving the formula.

theia-panel/src/scene/Scene.ts
```diff
@@ -85,7 +85,14 @@ export function createScene(container: HTMLElement): SceneContext {
       const rotateSpeed = 0.005;
       theta -= dxPixel * rotateSpeed;
       phi -= dyPixel * rotateSpeed;
-      phi = Math.max(0.01, Math.min(Math.PI - 0.01, phi));
+      if (phi < 0) {
+        phi = -phi;
+        theta += Math.PI;
+      } else if (phi > Math.PI) {
+        phi = 2 * Math.PI - phi;
```
zebster-cmd: **SUGGESTION**: The new pole-reflection logic only handles a single pole crossing per drag tick. A very fast vertical drag can produce |dyPixel * 0.005| > π, in which case `phi` lands outside [-π, 2π] and the reflection leaves it outside [0, π] before clamping snaps it to the pole. Not a correctness bug, but it produces a noticeable visual snap at the pole on flick gestures.

💡 Use modular arithmetic so any drag delta is handled uniformly, e.g.:

```ts
phi = ((phi % (2 * Math.PI)) + 2 * Math.PI) % (2 * Math.PI);
if (phi > Math.PI) { phi = 2 * Math.PI - phi; theta += Math.PI; }
```

theia-panel/src/index.ts
```diff
@@ -132,7 +132,7 @@ export async function mount(
     modelFilter?: string | null,
   ): Set<string> {
     const kindVisible = new Set<string>();
-    if (enabledKinds.has("subagent")) {
+    if (enabledKinds.has("subagent") || enabledKinds.has("cron-chain")) {
```
zebster-cmd: **SUGGESTION**: Expanding the "force all nodes visible" branch to include cron-chain mirrors the subagent special-case, but the rationale isn't stated and will become harder to reason about as more edge kinds are added. Worth a one-line comment naming the invariant ("sparse edge kinds that may be the only connection a node has → keep all nodes visible when enabled").

💡 Add a brief comment above the conditional explaining why these specific kinds bypass connectivity-based filtering, so the next added edge kind has clear guidance.

zebster-cmd (changes requested):
## Code Review

The cron-chain feature is well-integrated across detect/ingest/types/UI/physics layers and the additions follow existing theia patterns cleanly. However, there is a critical Python bug in detect/cron.py: `zip(group, group[1:], strict=True)` will raise ValueError on every cron job with 2+ sessions, since the iterables differ in length by exactly one. There's also a regression risk in install.sh where `git clone --branch` will not accept the commit SHAs the inline comment explicitly invites.

zebster-cmd:
## Code Review Summary

**Verdict:** REQUEST_CHANGES

The cron-chain feature is well-integrated across detect/ingest/types/UI/physics layers and the additions follow existing theia patterns cleanly. However, there is a critical Python bug in detect/cron.py: `zip(group, group[1:], strict=True)` will raise ValueError on every cron job with 2+ sessions, since the iterables differ in length by exactly one. There's also a regression risk in install.sh where `git clone --branch` will not accept the commit SHAs the inline comment explicitly invites.

### Findings
- 🔴 **theia-core/theia_core/detect/cron.py:24** — `strict=True` will raise ValueError every time this loop runs with a non-trivial group. `group` and `group[1:]` always differ in length by exactly one, so as soon as the shorter iterable is exhausted, strict-zip will detect the mismatch and raise. This means cron-chain edges will never actually be emitted (and worse, the build pipeline will crash whenever any cron job has ≥2 sessions).
- ⚠️ **install.sh:49** — Switching from tarball-download to `git clone --depth 1 --branch "$REPO_TAG"` regresses commit-SHA support. The inline comment on REPO_TAG still says `pin to a release tag or commit SHA for production`, but `git clone --branch` only accepts branch and tag refs — pinning REPO_TAG to a commit SHA will now fail with `Remote branch <sha> not found`. The previous tarball URL `archive/<sha>.tar.gz` worked for any ref. This also adds a hard dependency on `git` being installed on the user's machine for the curl-piped install path, where previously only curl+tar were needed.
- 💡 **theia-panel/src/ui/SidePanel.ts:288** — `ev.interval_hours` is typed `unknown` (evidence is `Record<string, unknown>`), so `Number(intervalHours).toFixed(1)` can render the string "NaN" if the field is ever missing or non-numeric. The `!== undefined` check doesn't catch null or non-numeric strings.
- 💡 **theia-core/theia_core/detect/cron.py:26** — The weight formula `1 / (1 + hours * 0.1)` is undocumented. A future reader (or you, in 6 months) won't know whether the 0.1 decay constant is intentional or a magic number — and at hours=0 (clock-skewed sessions or simultaneous runs) weight is 1.0, which equals the strongest semantic edges in the system. Consider whether that's intended.
- 💡 **theia-panel/src/scene/Scene.ts:92** — The new pole-reflection logic only handles a single pole crossing per drag tick. A very fast vertical drag can produce |dyPixel * 0.005| > π, in which case `phi` lands outside [-π, 2π] and the reflection leaves it outside [0, π] before clamping snaps it to the pole. Not a correctness bug, but it produces a noticeable visual snap at the pole on flick gestures.
- 💡 **theia-panel/src/index.ts:135** — Expanding the "force all nodes visible" branch to include cron-chain mirrors the subagent special-case, but the rationale isn't stated and will become harder to reason about as more edge kinds are added. Worth a one-line comment naming the invariant ("sparse edge kinds that may be the only connection a node has → keep all nodes visible when enabled").

zebster-cmd:
Rebuilt branch from `main` with only the curl-pipe README commit (originally `dd31ff2`). Earlier shared commits that belonged to PR #53/#54 have been removed from this branch. New tip: `9033f77`.

zebster-cmd:
Rebased on top of `main` (now at `3e3e0b0`). Conflict was caused by #56 having migrated the long-form Quick start / Prerequisites / Usage sections out of the README into `docs/quick-start.md` and `docs/usage.md`.

Adapted the change to the new README structure:

- Added a short `## Install` section between the intro and `## Documentation` containing only the curl-pipe one-liner (the PR's stated primary intent).
- Linked to `docs/quick-start.md` for alternative install methods, prerequisites, and configuration (those already include the one-liner verbatim from #56).

Net diff is now +10 lines on README.md only — no duplication with the docs/ tree. New tip: `dd454e3`.

PR #56, opened 2026-04-25
docs: clean up docs/ directory, migrate README content to standalone files

### LINKED ISSUE #52: clean up docs/ directory and move content out of README
## Description

- `docs/` contains junk / placeholder files instead of actual documentation
- The README is cluttered with what should be standalone docs pages

### Suggested fix

1. Clear out junk from `docs/`
2. Move detailed content from the README into proper files in `docs/`
3. Keep the README as a brief overview with links to `docs/`

arm64be:
Closes #52


theia-core/theia_core/detect/cron.py
```diff
@@ -0,0 +1,39 @@
...
+    """
+    by_job: dict[str, list[Session]] = defaultdict(list)
+    for sess in sessions:
+        if sess.cron_job_id:
+            by_job[sess.cron_job_id].append(sess)
+
+    edges: list[Edge] = []
+    for job_id, group in by_job.items():
+        group.sort(key=lambda s: s.started_at)
+        for prev_sess, next_sess in zip(group, group[1:], strict=True):
```
zebster-cmd: **CRITICAL**: `zip(group, group[1:], strict=True)` will raise ValueError for any cron job with ≥2 sessions. `group` and `group[1:]` always differ in length by 1 by construction, and `strict=True` requires equal length — so the detector crashes precisely when it has work to do. A 1-session group exits the inner loop without iterating and is fine; a 2+ session group (the feature's actual purpose) raises. Confirmed locally: `zip([1,2,3], [2,3], strict=True)` → ValueError.

💡 Drop `strict=True` (or use `itertools.pairwise(group)`):

    from itertools import pairwise
    for prev_sess, next_sess in pairwise(group):
        ...

install.sh
```diff
@@ -40,16 +40,14 @@ REPO_TAG="main"  # pin to a release tag or commit SHA for production
...
     TMPDIR="$(mktemp -d)"
-    ARCHIVE_URL="${REPO_URL}/archive/${REPO_TAG}.tar.gz"
 
     echo ""
-    echo "  Downloading Theia from ${ARCHIVE_URL} ..."
+    echo "  Cloning Theia from ${REPO_URL} ..."
     echo ""
 
-    curl -fsSL "$ARCHIVE_URL" | tar xz -C "$TMPDIR"
+    git clone --depth 1 --branch "$REPO_TAG" "$REPO_URL" "$TMPDIR"
```
zebster-cmd, line 53: **WARNING**: `git clone --depth 1 --branch "$REPO_TAG"` accepts a branch or tag, but NOT a commit SHA. The comment on line 33 says "pin to a release tag or commit SHA for production" — anyone following that advice will get `fatal: Remote branch <sha> not found`. Either drop the SHA option from the comment, or fall back to `git clone` followed by `git checkout $REPO_TAG` so SHAs work. Also note: the surrounding comment on lines 37–39 still says "Download a tarball" — stale after the switch to `git clone`.

💡 Either tighten the comment to "a release tag or branch name", or replace the clone with:

    git clone --depth 1 "$REPO_URL" "$TMPDIR"
    if [ "$REPO_TAG" != "main" ]; then
        git -C "$TMPDIR" fetch --depth 1 origin "$REPO_TAG"
        git -C "$TMPDIR" checkout FETCH_HEAD
    fi

And update the comment block to say `git clone` rather than tarball.

docs/quick-start.md
```diff
@@ -0,0 +1,62 @@
+# Quick Start
+
+The fastest way to install theia:
+
+```bash
+curl -fsSL https://raw.githubusercontent.com/arm64be/theia/main/install.sh | sh
```
zebster-cmd, line 6: **WARNING**: `curl ... | sh` is unreliable because install.sh's bootstrap section uses `${BASH_SOURCE[0]:-}` (bash array syntax). On systems where `/bin/sh` is dash (most Linux distros), this errors with "Bad substitution" before reaching the `exec bash` re-exec. install.sh's own header on line 37 already documents the correct invocation as `| bash`. docs/usage.md has the same issue.

💡 Change `| sh` to `| bash` in both docs/quick-start.md and docs/usage.md.

theia-panel/src/scene/Scene.ts
```diff
@@ -85,7 +85,14 @@ export function createScene(container: HTMLElement): SceneContext {
       const rotateSpeed = 0.005;
       theta -= dxPixel * rotateSpeed;
       phi -= dyPixel * rotateSpeed;
-      phi = Math.max(0.01, Math.min(Math.PI - 0.01, phi));
+      if (phi < 0) {
```
zebster-cmd: **WARNING**: The pitch-wrap logic only handles a single overflow. A fast/long mouse drag where `dyPixel * rotateSpeed` exceeds π (≈ 628px at rotateSpeed=0.005) leaves phi outside [0, π] after a single reflection, and the final clamp snaps it to a near-pole orientation that doesn't match user intent. For example, phi pre-wrap = -1.5π → after reflection phi = 1.5π → elif doesn't re-trigger → clamped to π-0.001.

💡 Normalize phi modulo 2π first, then reflect:

    phi = ((phi % (2 * Math.PI)) + 2 * Math.PI) % (2 * Math.PI);
    if (phi > Math.PI) {
      phi = 2 * Math.PI - phi;
      theta += Math.PI;
    }
    phi = Math.max(0.001, Math.min(Math.PI - 0.001, phi));

theia-panel/src/index.ts
```diff
@@ -132,7 +132,7 @@ export async function mount(
     modelFilter?: string | null,
   ): Set<string> {
     const kindVisible = new Set<string>();
-    if (enabledKinds.has("subagent")) {
+    if (enabledKinds.has("subagent") || enabledKinds.has("cron-chain")) {
```
zebster-cmd: **SUGGESTION**: Adding `cron-chain` to the same branch as `subagent` short-circuits the entire visibility filter: when cron-chain is enabled, every node in the graph is marked visible. The original `subagent` shortcut existed because subagent leaf nodes are otherwise hidden by the parent_id check below. Cron-spawned sessions don't have `parent_id` (they're top-level), so they're already visible via the `else` branch — adding them here probably overshoots intent and could mask other filters. Worth verifying with a graph that has cron-chain edges plus disabled memory-share/cross-search.

💡 Drop `|| enabledKinds.has("cron-chain")` and let cron-chain go through the same `enabledKinds.has(edge.kind)` path as memory-share/cross-search/tool-overlap. If there's a specific node-visibility issue with cron sessions, handle it explicitly rather than reusing the subagent shortcut.

theia-panel/src/data/types.ts
```diff
@@ -33,7 +33,7 @@ export interface TheiaGraph {
   edges: {
     source: string;
     target: string;
-    kind: "memory-share" | "cross-search" | "tool-overlap" | "subagent";
+    kind: "memory-share" | "cross-search" | "tool-overlap" | "subagent" | "cron-chain";
```
zebster-cmd: **SUGGESTION**: types.ts is generated by `npm run generate-types` from schemas/graph.schema.json (per package.json). Confirm this diff is the result of regeneration, not a hand-edit — otherwise the next CI `generate-types` run will revert this change. (The output looks correct, just flagging the workflow.)

💡 Run `npm run generate-types` if not already done, and ensure CI's panel.yml `generate-types` step produces a clean diff.

zebster-cmd (changes requested):
## Code Review

The cron-chain detector contains a fatal bug: `zip(group, group[1:], strict=True)` raises ValueError whenever a cron job has 2+ sessions — i.e., every case the feature is meant to handle. There are also several smaller issues: the install bootstrap docs/code drift (comment says "tarball" but code clones; `git clone --branch` rejects commit SHAs even though the inline comment recommends them), `curl | sh` in quick-start.md is unreliable for a script using bash arrays, and the visibility-filter shortcut for `cron-chain` likely has different intent than `subagent`. Once the strict=True bug is fixed, the rest is approachable as follow-up.

zebster-cmd:
## Code Review Summary

**Verdict:** REQUEST_CHANGES

The cron-chain detector contains a fatal bug: `zip(group, group[1:], strict=True)` raises ValueError whenever a cron job has 2+ sessions — i.e., every case the feature is meant to handle. There are also several smaller issues: the install bootstrap docs/code drift (comment says "tarball" but code clones; `git clone --branch` rejects commit SHAs even though the inline comment recommends them), `curl | sh` in quick-start.md is unreliable for a script using bash arrays, and the visibility-filter shortcut for `cron-chain` likely has different intent than `subagent`. Once the strict=True bug is fixed, the rest is approachable as follow-up.

### Findings
- 🔴 **theia-core/theia_core/detect/cron.py:24** — `zip(group, group[1:], strict=True)` will raise ValueError for any cron job with ≥2 sessions. `group` and `group[1:]` always differ in length by 1 by construction, and `strict=True` requires equal length — so the detector crashes precisely when it has work to do. A 1-session group exits the inner loop without iterating and is fine; a 2+ session group (the feature's actual purpose) raises. Confirmed locally: `zip([1,2,3], [2,3], strict=True)` → ValueError.
- ⚠️ **install.sh:48** — `git clone --depth 1 --branch "$REPO_TAG"` accepts a branch or tag, but NOT a commit SHA. The comment on line 33 says "pin to a release tag or commit SHA for production" — anyone following that advice will get `fatal: Remote branch <sha> not found`. Either drop the SHA option from the comment, or fall back to `git clone` followed by `git checkout $REPO_TAG` so SHAs work. Also note: the surrounding comment on lines 37–39 still says "Download a tarball" — stale after the switch to `git clone`.
- ⚠️ **docs/quick-start.md:6** — `curl ... | sh` is unreliable because install.sh's bootstrap section uses `${BASH_SOURCE[0]:-}` (bash array syntax). On systems where `/bin/sh` is dash (most Linux distros), this errors with "Bad substitution" before reaching the `exec bash` re-exec. install.sh's own header on line 37 already documents the correct invocation as `| bash`. docs/usage.md has the same issue.
- ⚠️ **theia-panel/src/scene/Scene.ts:88** — The pitch-wrap logic only handles a single overflow. A fast/long mouse drag where `dyPixel * rotateSpeed` exceeds π (≈ 628px at rotateSpeed=0.005) leaves phi outside [0, π] after a single reflection, and the final clamp snaps it to a near-pole orientation that doesn't match user intent. For example, phi pre-wrap = -1.5π → after reflection phi = 1.5π → elif doesn't re-trigger → clamped to π-0.001.
- 💡 **theia-panel/src/index.ts:135** — Adding `cron-chain` to the same branch as `subagent` short-circuits the entire visibility filter: when cron-chain is enabled, every node in the graph is marked visible. The original `subagent` shortcut existed because subagent leaf nodes are otherwise hidden by the parent_id check below. Cron-spawned sessions don't have `parent_id` (they're top-level), so they're already visible via the `else` branch — adding them here probably overshoots intent and could mask other filters. Worth verifying with a graph that has cron-chain edges plus disabled memory-share/cross-search.
- 💡 **theia-panel/src/data/types.ts:36** — types.ts is generated by `npm run generate-types` from schemas/graph.schema.json (per package.json). Confirm this diff is the result of regeneration, not a hand-edit — otherwise the next CI `generate-types` run will revert this change. (The output looks correct, just flagging the workflow.)

zebster-cmd:
CI fixes pushed (same patterns as PR #54): itertools.pairwise for ruff B905 and prettier on data/types.ts + ui/SidePanel.ts. New tip: b4f3d5c4598be7ba3f9181fca924be6a08945579.

zebster-cmd:
Rebuilt branch from `main` with only the docs-cleanup commit (originally `9639049`). Earlier shared commits that belonged to PR #53/#54 have been removed from this branch. New tip: `3a75e85`.

Note: README.md ended up resolved by taking the cherry-picked commit's end-state, so the README on this branch reflects the docs-cleanup intent. Review the README diff to confirm it's what you want for this PR scope (independent of #55).

Issue #57, opened 2026-04-25
bug(plugin): constellation tab does not follow Hermes dashboard viewport conventions


zebster-cmd:
## Bug Report

The Constellation tab does not follow the Hermes dashboard's page layout conventions. PR #30's responsive overlay approach (absolute-positioned filter bar, search bar, and side panel inside the iframe) conflicts with the dashboard's native page header and viewport model, causing truncation, overlap, and cramped visual hierarchy.

### What the dashboard conventions look like

Every built-in tab (Sessions, Analytics, Logs, Config, etc.) follows this structure via `PageHeaderProvider`:

```
┌───────────────────────────────────────────────────────────────────┐
│  [Title] [badge] [badge]                    [search/actions]     │  ← Page header (h-14, fixed)
├───────────────────────────────────────────────────────────────────┤
│                                                                    │
│                          Main content                              │  ← Main area (flex-1, scrollable)
│                     (fills remaining height)                        │
│                                                                    │
└───────────────────────────────────────────────────────────────────┘
```

Pages use `usePageHeader()` to inject into the header:
- `afterTitle` — badges, stats, version
- `end` — search input, action buttons, refresh controls

The `<main>` area then gets 100% of the remaining viewport height with `flex-1`.

### What the Constellation tab currently does (wrong)

The plugin renders its **own** `Card` header inside the page content:

```
┌───────────────────────────────────────────────────────────────────┐
│  Constellation                                                       │  ← Dashboard page header (empty!)
├───────────────────────────────────────────────────────────────────┤
│  ┌──────────────────────────────────────────────────────────┐   │
│  │ Session Constellation  v0.1.0  16 sessions / 121 edges  [Reload]  │   │  ← Plugin's own Card header
│  └──────────────────────────────────────────────────────────┘   │
│                                                                    │
│  [☑] Memory Share  [☑] Cross Search ...    [Search sessions...]     │  ← Absolute overlays (filter + search)
│                                                                    │
│          ●───●───●                                                │
│           \───●───●              ┌───────────────┐               │
│             ●───●                │  Side panel  │               │  ← Side panel slides over graph
│               ●                   │  (45vw)      │               │
│                                    └───────────────┘               │
└───────────────────────────────────────────────────────────────────┘
```

This creates multiple problems:

1. **Double header** — The dashboard renders an empty "Constellation" page header, then the plugin adds its own title card below it. Title is duplicated, vertical space is wasted.

[...]

PR #67, opened 2026-04-25
fix(panel): polished UI — collapsed filter, search minimize, overlay blocking, focus filter, localStorage persistence

### LINKED ISSUE #27: bug(panel): search bar, filter bar, and side panel overlap on narrow viewports

arm64be:
## Summary

Fixes #27 (narrow viewport overlap) and #57 (dashboard convention violations) with a series of UI refinements.

### Filter bar collapsed to dropdown
- Filter bar now always shows as a compact **"Filters"** button (no inline toggle row)
- Clicking opens a dropdown with toggle switches and model select in a vertical column
- Button highlights (darker background, accent border) when dropdown is open
- Click-outside closes the dropdown

### Search bar minimizes when panel is open
- When the side panel is open, the search bar hides the full input and shows a small **⌕ icon button** next to the Filters button
- Clicking the icon closes the side panel and restores the normal centered search bar with input focused
- Both buttons use identical font/padding so the bar height stays stable

### Pointer events blocked on all overlays
- `data-ui-overlay` attribute added to filter bar, search bar, side panel (`<aside>`)
- Picker hover detection, mousedown (drag start), and mouseup (node click) all check for the attribute
- Nodes behind any UI overlay can no longer be hovered, clicked, or dragged through

### Focus mode picking fix
- Added `focusFilter` state variable tracked alongside `visibleNodeIds`
- Picker's `isVisible` callback checks both, preventing selection of focus-filtered nodes
- Cleared on filter changes and focus toggle

### Visual consistency
- All transitions synced to **100ms** (matching Hermes dashboard convention)
- Removed `outline` on search toggle button

### Filter state persistence
- Edge kind toggles and model selection saved to `localStorage` under `theia-constellation-filter`
- Restored on page load — filters survive refreshes

## Files changed
| File | Changes |
|---|---|
| `theia-panel/src/index.ts` | localStorage helpers, focusFilter, overlay checks, search/filter wiring |
| `theia-panel/src/ui/FilterBar.ts` | Collapsed dropdown, search toggle button, `onSearchToggle` callback |
| `theia-panel/src/ui/SearchBar.ts` | `setPanelOpen` for minimize/restore behavior |
| `theia-panel/src/ui/SidePanel.ts` | Added `data-ui-overlay` attribute |
| `theia-panel/src/scene/Picker.ts` | Added `[data-ui-overlay]` check in hover detection |

Closes #27, #57

### DIFF
--- a/theia-panel/src/index.ts
+++ b/theia-panel/src/index.ts
@@ -25,6 +25,14 @@ export interface Controller {
   reload(graphUrl?: string): Promise<void>;
 }
 
+const VALID_KINDS: TheiaGraph["edges"][number]["kind"][] = [
+  "memory-share",
+  "cross-search",
+  "tool-overlap",
+  "subagent",
+  "cron-chain",
+];
+
 const DEFAULT_KINDS: TheiaGraph["edges"][number]["kind"][] = [
   "memory-share",
   "cross-search",
@@ -40,9 +48,14 @@ function loadFilterState(): {
     const raw = localStorage.getItem(STORAGE_KEY);
     if (!raw) return null;
     const parsed = JSON.parse(raw);
+    const parsedKinds = Array.isArray(parsed?.kinds)
+      ? parsed.kinds.filter((k: unknown): k is TheiaGraph["edges"][number]["kind"] =>
+          (VALID_KINDS as readonly unknown[]).includes(k),
+        )
+      : DEFAULT_KINDS;
     return {
-      kinds: new Set(parsed.kinds ?? DEFAULT_KINDS),
-      model: parsed.model ?? null,
+      kinds: new Set(parsedKinds.length > 0 ? parsedKinds : DEFAULT_KINDS),
+      model: typeof parsed?.model === "string" ? parsed.model : null,
     };
   } catch {
     return null;
@@ -125,6 +138,12 @@ export async function mount(
     nodes.setHighlight(idx, false);
   }
 
+  function enterPanelMode(node: TheiaGraph["nodes"][number], related: TheiaGraph["edges"]) {
+    sidePanel.show(node, related);
+    searchBar.setPanelOpen(true);
+    filterBar.setSearchToggleVisible(true);
+  }
+
   const sidePanel = createSidePanel(element, theme, {
     onNavigate: (targetId) => {
       const idx = nodeIndex.get(targetId);
@@ -137,9 +156,7 @@ export async function mount(
       const related = currentGraph.edges.filter(
         (e) => (e.source === n.id || e.target === n.id) && kinds.has(e.kind),
       );
-      sidePanel.show(n, related);
-      searchBar.setPanelOpen(true);
-      filterBar.setSearchToggleVisible(true);
+      enterPanelMode(n, related);
       applyFocusModeIfEnabled(n.id);
       emit("node-click", targetId);
     },
@@ -152,6 +169,7 @@ export async function mount(
     onFocusToggle: (enabled) => {
       focusEnabled = enabled;
       if (!focusEnabled) {
+        focusFilter = null;
         updateVisibility();
       } else {
         const id = sidePanel.currentNodeId();
@@ -217,7 +235,6 @@ export async function mount(
   let visibleNodeIds = new Set<string>();
 
   function updateVisibility() {
-    focusFilter = null;
     visibleNodeIds = computeVisibleNodeIds(currentGraph, kinds, modelFilter);
     for (let i = 0; i < currentGraph.nodes.length; i++) {
       nodes.setVisible(i, visibleNodeIds.has(currentGraph.nodes[i]!.id));
@@ -372,9 +389,7 @@ export async function mount(
             (e.source === result.node.id || e.target === result.node.id) &&
             kinds.has(e.kind),
         );
-        sidePanel.show(result.node, related);
-        searchBar.setPanelOpen(true);
-        filterBar.setSearchToggleVisible(true);
+        enterPanelMode(result.node, related);
       },
       theme,
       (node) => visibleNodeIds.has(node.id),
@@ -494,9 +509,7 @@ export async function mount(
         const related = currentGraph.edges.filter(
           (e) => (e.source === n.id || e.target === n.id) && kinds.has(e.kind),
         );
-        sidePanel.show(n, related);
-        searchBar.setPanelOpen(true);
-        filterBar.setSearchToggleVisible(true);
+        enterPanelMode(n, related);
         applyFocusModeIfEnabled(n.id);
         emit("node-click", n.id);
       } else {
@@ -535,6 +548,7 @@ export async function mount(
       kinds = state.kinds;
       modelFilter = state.model;
       saveFilterState(kinds, modelFilter);
+      focusFilter = null;
       updateVisibility();
       const id = sidePanel.currentNodeId();
       if (id) applyFocusModeIfEnabled(id);
@@ -614,9 +628,7 @@ export async function mount(
             (e) =>
               (e.source === n.id || e.target === n.id) && kinds.has(e.kind),
           );
-          sidePanel.show(n, related);
-          searchBar.setPanelOpen(true);
-          filterBar.setSearchToggleVisible(true);
+          enterPanelMode(n, related);
           applyFocusModeIfEnabled(n.id);
         }
       }
--- a/theia-panel/src/index.ts
+++ b/theia-panel/src/index.ts
@@ -30,6 +30,36 @@ const DEFAULT_KINDS: TheiaGraph["edges"][number]["kind"][] = [
   "cross-search",
 ];
 
+const STORAGE_KEY = "theia-constellation-filter";
+
+function loadFilterState(): {
+  kinds: Set<TheiaGraph["edges"][number]["kind"]>;
+  model: string | null;
+} | null {
+  try {
+    const raw = localStorage.getItem(STORAGE_KEY);
+    if (!raw) return null;
+    const parsed = JSON.parse(raw);
+    return {
+      kinds: new Set(parsed.kinds ?? DEFAULT_KINDS),
+      model: parsed.model ?? null,
+    };
+  } catch {
+    return null;
+  }
+}
+
+function saveFilterState(kinds: Set<string>, model: string | null): void {
+  try {
+    localStorage.setItem(
+      STORAGE_KEY,
+      JSON.stringify({ kinds: Array.from(kinds), model }),
+    );
+  } catch {
+    /* quota exceeded, ignore */
+  }
+}
+
 export async function mount(
   element: HTMLElement,
   graphUrl: string,
@@ -60,6 +90,11 @@ export async function mount(
 
   let kinds = new Set(options.edgeKinds ?? DEFAULT_KINDS);
   let modelFilter: string | null = null;
+  const saved = loadFilterState();
+  if (saved) {
+    kinds = saved.kinds;
+    modelFilter = saved.model;
+  }
   let focusEnabled = false;
   let focusFilter: Set<string> | null = null;
 
@@ -498,11 +533,13 @@ export async function mount(
     (state) => {
       kinds = state.kinds;
       modelFilter = state.model;
+      saveFilterState(kinds, modelFilter);
       updateVisibility();
       const id = sidePanel.currentNodeId();
       if (id) applyFocusModeIfEnabled(id);
     },
     theme,
+    modelFilter,
     () => {
       sidePanel.hide();
       searchBar.input.focus();
--- a/theia-panel/src/index.ts
+++ b/theia-panel/src/index.ts
@@ -103,11 +103,15 @@ export async function mount(
         (e) => (e.source === n.id || e.target === n.id) && kinds.has(e.kind),
       );
       sidePanel.show(n, related);
+      searchBar.setPanelOpen(true);
+      filterBar.setSearchToggleVisible(true);
       applyFocusModeIfEnabled(n.id);
       emit("node-click", targetId);
     },
     onClose: () => {
       clearSelected();
+      searchBar.setPanelOpen(false);
+      filterBar.setSearchToggleVisible(false);
       nodes.flush();
     },
     onFocusToggle: (enabled) => {
@@ -334,6 +338,8 @@ export async function mount(
             kinds.has(e.kind),
         );
         sidePanel.show(result.node, related);
+        searchBar.setPanelOpen(true);
+        filterBar.setSearchToggleVisible(true);
       },
       theme,
       (node) => visibleNodeIds.has(node.id),
@@ -454,6 +460,8 @@ export async function mount(
           (e) => (e.source === n.id || e.target === n.id) && kinds.has(e.kind),
         );
         sidePanel.show(n, related);
+        searchBar.setPanelOpen(true);
+        filterBar.setSearchToggleVisible(true);
         applyFocusModeIfEnabled(n.id);
         emit("node-click", n.id);
       } else {
@@ -495,6 +503,10 @@ export async function mount(
       if (id) applyFocusModeIfEnabled(id);
     },
     theme,
+    () => {
+      sidePanel.hide();
+      searchBar.input.focus();
+    },
   );
 
   const listeners: Record<string, Array<(...args: unknown[]) => void>> = {
@@ -565,6 +577,8 @@ export async function mount(
               (e.source === n.id || e.target === n.id) && kinds.has(e.kind),
           );
           sidePanel.show(n, related);
+          searchBar.setPanelOpen(true);
+          filterBar.setSearchToggleVisible(true);
           applyFocusModeIfEnabled(n.id);
         }
       }
--- a/theia-panel/src/index.ts
+++ b/theia-panel/src/index.ts
@@ -49,8 +49,9 @@ function loadFilterState(): {
     if (!raw) return null;
     const parsed = JSON.parse(raw);
     const parsedKinds = Array.isArray(parsed?.kinds)
-      ? parsed.kinds.filter((k: unknown): k is TheiaGraph["edges"][number]["kind"] =>
-          (VALID_KINDS as readonly unknown[]).includes(k),
+      ? parsed.kinds.filter(
+          (k: unknown): k is TheiaGraph["edges"][number]["kind"] =>
+            (VALID_KINDS as readonly unknown[]).includes(k),
         )
       : DEFAULT_KINDS;
     return {
@@ -138,7 +139,10 @@ export async function mount(
     nodes.setHighlight(idx, false);
   }
 
-  function enterPanelMode(node: TheiaGraph["nodes"][number], related: TheiaGraph["edges"]) {
+  function enterPanelMode(
+    node: TheiaGraph["nodes"][number],
+    related: TheiaGraph["edges"],
+  ) {
     sidePanel.show(node, related);
     searchBar.setPanelOpen(true);
     filterBar.setSearchToggleVisible(true);
--- a/theia-panel/src/index.ts
+++ b/theia-panel/src/index.ts
@@ -61,6 +61,7 @@ export async function mount(
   let kinds = new Set(options.edgeKinds ?? DEFAULT_KINDS);
   let modelFilter: string | null = null;
   let focusEnabled = false;
+  let focusFilter: Set<string> | null = null;
 
   // Mutable graph-specific state — closures capture the binding, not the value
   let currentGraph: TheiaGraph;
@@ -177,6 +178,7 @@ export async function mount(
   let visibleNodeIds = new Set<string>();
 
   function updateVisibility() {
+    focusFilter = null;
     visibleNodeIds = computeVisibleNodeIds(currentGraph, kinds, modelFilter);
     for (let i = 0; i < currentGraph.nodes.length; i++) {
       nodes.setVisible(i, visibleNodeIds.has(currentGraph.nodes[i]!.id));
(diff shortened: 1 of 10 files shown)


theia-panel/src/ui/FilterBar.ts
```diff
@@ -57,25 +57,88 @@ export function createFilterBar(
...
+      min-width: 190px;
+      display: ${dropdownOpen ? "block" : "none"};
+    `;
+    content.style.cssText = `
+      display: flex; flex-direction: column; gap: 8px;
+      font: 10px/1.4 var(--theia-font, ${FONT_STACK});
+      letter-spacing: 0.05em;
+      color: #${theme.fg};
+    `;
+    for (const child of content.children) {
```
zebster-cmd, line 132: **CRITICAL**: Inside applyBarStyle(), the loop overwrites every kind label's cssText with a fixed string that omits the per-kind `color: ${state.has(kind) ? fg : fg2}` rule. Because applyBarStyle is now called after the labels are appended (and re-runs on every dropdown toggle / search-toggle visibility change / theme change), unselected kinds will no longer be dimmed — they all render in the default fg color until the user hovers (which restores the original behavior via onmouseenter/leave). This is a real visual regression for the primary affordance of the filter UI.

💡 Either drop this loop (the for-loop in the kinds setup already sets the correct cssText including state-aware color), or recompute the color per label here. Cleanest: keep a `(label, kind)` map and have applyBarStyle iterate it, writing `color: ${state.has(kind) ? '#'+theme.fg : '#'+theme.fg2}` along with the layout properties.

theia-panel/src/index.ts
```diff
@@ -478,11 +533,17 @@ export async function mount(
...
       modelFilter = state.model;
+      saveFilterState(kinds, modelFilter);
       updateVisibility();
       const id = sidePanel.currentNodeId();
       if (id) applyFocusModeIfEnabled(id);
     },
     theme,
+    modelFilter,
+    () => {
+      sidePanel.hide();
```
zebster-cmd, line 563: **WARNING**: The new onSearchToggle callback calls `sidePanel.hide()` then `searchBar.input.focus()`, but does not call `searchBar.setPanelOpen(false)` / `filterBar.setSearchToggleVisible(false)`. Unless `sidePanel.hide()` invokes the `onClose` callback that resets those flags, the search bar remains `display: none` and `.focus()` on a hidden input is a no-op — so clicking the ⌕ button hides the panel but fails to surface and focus the search field.

💡 Either ensure sidePanel.hide() routes through the same teardown as onClose (preferred — single source of truth), or explicitly call `searchBar.setPanelOpen(false); filterBar.setSearchToggleVisible(false);` here before focusing.

theia-panel/src/index.ts
```diff
@@ -102,11 +138,15 @@ export async function mount(
         (e) => (e.source === n.id || e.target === n.id) && kinds.has(e.kind),
       );
       sidePanel.show(n, related);
+      searchBar.setPanelOpen(true);
```
zebster-cmd: **WARNING**: `searchBar.setPanelOpen(true)` and `filterBar.setSearchToggleVisible(true)` are duplicated in four show-paths (sidePanel.onSelect, search-result onSelect, mouseup click handler, and the public select() API at ~line 617). Any future show path that forgets the pair will leave the search bar visible while the side panel is open. The mirror cleanup is centralized in onClose only.

💡 Encapsulate the side-effect inside SidePanel.show()/hide() (and have SidePanel emit/own the dependency on searchBar+filterBar), or wrap the trio in a local `openSidePanel(node, related)` helper used by all four call sites.

theia-panel/src/index.ts
```diff
@@ -30,6 +30,36 @@ const DEFAULT_KINDS: TheiaGraph["edges"][number]["kind"][] = [
...
+function loadFilterState(): {
+  kinds: Set<TheiaGraph["edges"][number]["kind"]>;
+  model: string | null;
+} | null {
+  try {
+    const raw = localStorage.getItem(STORAGE_KEY);
+    if (!raw) return null;
+    const parsed = JSON.parse(raw);
+    return {
+      kinds: new Set(parsed.kinds ?? DEFAULT_KINDS),
```
zebster-cmd: **WARNING**: loadFilterState() does not validate `parsed.kinds` is an array of valid edge kinds. If a user (or stale schema) writes a non-array (e.g. `null`-prototype object, number, or array of unknown strings), `new Set(parsed.kinds ?? DEFAULT_KINDS)` either throws (caught — falls back to null/no state) or silently constructs a Set with bogus kind strings that pass through as `TheiaGraph["edges"][number]["kind"]` via the function return type. The latter case yields a filter with zero matches and no obvious failure mode for the user.

💡 Validate before constructing: `const validKinds = Array.isArray(parsed.kinds) ? parsed.kinds.filter((k): k is TheiaGraph['edges'][number]['kind'] => DEFAULT_KINDS.includes(k)) : DEFAULT_KINDS;` and use that. Same for `parsed.model` (typeof === 'string' || === null).

theia-panel/src/ui/FilterBar.ts
```diff
@@ -57,25 +57,88 @@ export function createFilterBar(
...
   initialTheme: ThemeTokens,
+  initialModel?: string | null,
+  onSearchToggle?: () => void,
 ) {
   let theme = initialTheme;
   const bar = document.createElement("div");
+  bar.dataset.uiOverlay = "";
+  const content = document.createElement("div");
   const toggles: Array<HTMLButtonElement & { _applyStyle: () => void }> = [];
   let select: HTMLSelectElement | null = null;
```
zebster-cmd, line 68: **WARNING**: `let separator: HTMLSpanElement | null = null;` is now dead — the only assignments and the `bar.append(separator)` were removed when the separator UI was dropped. Will trip noUnusedLocals once anyone runs strict TS checks here.

💡 Delete the `separator` declaration and any remaining cleanup code that referenced it in rebuildModelSelect().

theia-panel/src/index.ts
```diff
@@ -478,11 +533,17 @@ export async function mount(
     (state) => {
       kinds = state.kinds;
       modelFilter = state.model;
+      saveFilterState(kinds, modelFilter);
       updateVisibility();
       const id = sidePanel.currentNodeId();
       if (id) applyFocusModeIfEnabled(id);
     },
```
zebster-cmd, line 559: **WARNING**: `saveFilterState(kinds, modelFilter)` is called only from the FilterBar onChange callback, but `kinds` and `modelFilter` are also seeded from localStorage on mount (via `loadFilterState`). If the persisted state had values the FilterBar dropped (e.g. removed kind), they remain in `kinds` (the closure variable used by the picker/visibility) but won't be reflected in the FilterBar UI until the user toggles something — silent UI/state divergence.

💡 After loading saved state, pass the loaded model into createFilterBar (already done via `modelFilter` arg) AND also pass the loaded kinds as `initial`/`options.edgeKinds` so the toggles render in sync with the runtime state. Currently `kinds` is overwritten after `new Set(options.edgeKinds ?? DEFAULT_KINDS)` but the FilterBar receives the original `initial` separately.

theia-panel/src/ui/SearchBar.ts
```diff
@@ -226,6 +228,12 @@ export function createSearchBar(
     }
   });
 
+  function setPanelOpen(open: boolean) {
+    panelOpen = open;
+    wrapper.style.display = panelOpen ? "none" : "";
```
zebster-cmd: **SUGGESTION**: Setting `wrapper.style.display = panelOpen ? 'none' : ''` works because applyWrapperStyle's cssText template includes `display: flex`, but it relies on that invariant. If applyWrapperStyle ever stops setting display, reopening the bar will leave it as `display: ''` (i.e. inline default for div = block) and break the layout.

💡 Either add a `hidden` boolean to applyWrapperStyle and have it bake the correct `display` into cssText, or use the `hidden` attribute / a CSS class to toggle visibility so the inline cssText stays single-source-of-truth.

theia-panel/src/ui/FilterBar.ts
```diff
@@ -122,9 +185,29 @@ export function createFilterBar(
...
+  };
+
+  searchToggle.onclick = () => onSearchToggle?.();
+
+  function onDocumentClick(e: MouseEvent) {
+    if (dropdownOpen && !bar.contains(e.target as Node)) {
+      closeDropdown();
+    }
+  }
+  document.addEventListener("click", onDocumentClick);
```
zebster-cmd: **SUGGESTION**: `document.addEventListener('click', onDocumentClick)` is correctly removed in dispose, but the listener fires for every click on the page and then does an O(n) `bar.contains(...)` check — fine for one panel, but if multiple constellations ever mount (or HMR re-mounts without cleanup) listeners accumulate. Consider using `pointerdown` and adding it only while `dropdownOpen` is true, then removing on close.

💡 In btn.onclick, `if (dropdownOpen) document.addEventListener('pointerdown', onDocumentClick); else document.removeEventListener('pointerdown', onDocumentClick);` — keeps the global handler scoped to when it's actually needed.

theia-panel/src/ui/SidePanel.ts
```diff
@@ -89,7 +90,7 @@ export function createSidePanel(
       position: absolute; top: 0; right: 0; bottom: 0; width: ${panelWidth};
       background: ${themeBgAlpha(theme, 0.95)}; border-left: 1px solid #${theme.border};
       color: #${theme.fg}; font: 13px/1.6 var(--theia-font, ${FONT_STACK});
-      transform: translateX(${currentId ? "0" : "100%"}); transition: transform 220ms ease-out;
+      transform: translateX(${currentId ? "0" : "100%"}); transition: transform 100ms ease-out;
```
zebster-cmd, line 93: **SUGGESTION**: Reducing the slide-in transition from 220ms → 100ms (also done in SearchBar 220ms→100ms and FilterBar toggles 150ms→100ms) is a meaningful UX shift. 100ms is below the threshold where users perceive motion as deliberate; for a panel of this width it can feel snappy-bordering-on-jarring vs. the previous 220ms.

💡 Consider 150–180ms as a middle ground, or keep 220ms for the SidePanel slide (large element) while shortening only the small toggle/border transitions to 100ms.

theia-panel/src/index.ts
```diff
@@ -177,6 +217,7 @@ export async function mount(
   let visibleNodeIds = new Set<string>();
 
   function updateVisibility() {
+    focusFilter = null;
```
zebster-cmd: **SUGGESTION**: `updateVisibility()` now unconditionally clears `focusFilter = null`, but downstream `applyFocusModeIfEnabled(id)` is only called from the FilterBar onChange path. Other callers of updateVisibility (graph reload, theme apply, etc.) will silently drop the focus filter even when focus is enabled and a node is selected.

💡 After `updateVisibility()` returns, re-apply focus if active: pull the `if (id) applyFocusModeIfEnabled(id)` block into a `refreshVisibility()` wrapper used by all callers, instead of only the onChange path.

zebster-cmd (changes requested):
## Code Review

zebster-cmd:
## Code Review Summary

**Verdict:** REQUEST_CHANGES



### Findings
- 🔴 **theia-panel/src/ui/FilterBar.ts:133** — Inside applyBarStyle(), the loop overwrites every kind label's cssText with a fixed string that omits the per-kind `color: ${state.has(kind) ? fg : fg2}` rule. Because applyBarStyle is now called after the labels are appended (and re-runs on every dropdown toggle / search-toggle visibility change / theme change), unselected kinds will no longer be dimmed — they all render in the default fg color until the user hovers (which restores the original behavior via onmouseenter/leave). This is a real visual regression for the primary affordance of the filter UI.
- ⚠️ **theia-panel/src/index.ts:544** — The new onSearchToggle callback calls `sidePanel.hide()` then `searchBar.input.focus()`, but does not call `searchBar.setPanelOpen(false)` / `filterBar.setSearchToggleVisible(false)`. Unless `sidePanel.hide()` invokes the `onClose` callback that resets those flags, the search bar remains `display: none` and `.focus()` on a hidden input is a no-op — so clicking the ⌕ button hides the panel but fails to surface and focus the search field.
- ⚠️ **theia-panel/src/index.ts:141** — `searchBar.setPanelOpen(true)` and `filterBar.setSearchToggleVisible(true)` are duplicated in four show-paths (sidePanel.onSelect, search-result onSelect, mouseup click handler, and the public select() API at ~line 617). Any future show path that forgets the pair will leave the search bar visible while the side panel is open. The mirror cleanup is centralized in onClose only.
- ⚠️ **theia-panel/src/index.ts:44** — loadFilterState() does not validate `parsed.kinds` is an array of valid edge kinds. If a user (or stale schema) writes a non-array (e.g. `null`-prototype object, number, or array of unknown strings), `new Set(parsed.kinds ?? DEFAULT_KINDS)` either throws (caught — falls back to null/no state) or silently constructs a Set with bogus kind strings that pass through as `TheiaGraph["edges"][number]["kind"]` via the function return type. The latter case yields a filter with zero matches and no obvious failure mode for the user.
- ⚠️ **theia-panel/src/ui/FilterBar.ts:68** — `let separator: HTMLSpanElement | null = null;` is now dead — the only assignments and the `bar.append(separator)` were removed when the separator UI was dropped. Will trip noUnusedLocals once anyone runs strict TS checks here.
- ⚠️ **theia-panel/src/index.ts:540** — `saveFilterState(kinds, modelFilter)` is called only from the FilterBar onChange callback, but `kinds` and `modelFilter` are also seeded from localStorage on mount (via `loadFilterState`). If the persisted state had values the FilterBar dropped (e.g. removed kind), they remain in `kinds` (the closure variable used by the picker/visibility) but won't be reflected in the FilterBar UI until the user toggles something — silent UI/state divergence.
- 💡 **theia-panel/src/ui/SearchBar.ts:233** — Setting `wrapper.style.display = panelOpen ? 'none' : ''` works because applyWrapperStyle's cssText template includes `display: flex`, but it relies on that invariant. If applyWrapperStyle ever stops setting display, reopening the bar will leave it as `display: ''` (i.e. inline default for div = block) and break the layout.
- 💡 **theia-panel/src/ui/FilterBar.ts:209** — `document.addEventListener('click', onDocumentClick)` is correctly removed in dispose, but the listener fires for every click on the page and then does an O(n) `bar.contains(...)` check — fine for one panel, but if multiple constellations ever mount (or HMR re-mounts without cleanup) listeners accumulate. Consider using `pointerdown` and adding it only while `dropdownOpen` is true, then removing on close.
- 💡 **theia-panel/src/ui/SidePanel.ts:93** — Reducing the slide-in transition from 220ms → 100ms (also done in SearchBar 220ms→100ms and FilterBar toggles 150ms→100ms) is a meaningful UX shift. 100ms is below the threshold where users perceive motion as deliberate; for a panel of this width it can feel snappy-bordering-on-jarring vs. the previous 220ms.
- 💡 **theia-panel/src/index.ts:220** — `updateVisibility()` now unconditionally clears `focusFilter = null`, but downstream `applyFocusModeIfEnabled(id)` is only called from the FilterBar onChange path. Other callers of updateVisibility (graph reload, theme apply, etc.) will silently drop the focus filter even when focus is enabled and a node is selected.

arm64be:
Note: This PR also includes a fix for the plugin API import (replaced relative import with sys.path) which was out of original scope but was blocking dev testing — the relative import `from .graph_data import load_graph` breaks when Hermes loads the module via `importlib.util.spec_from_file_location()` (no package context).

arm64be:
## Unresolved: iframe content appears 125% scaled in dashboard

The graph renders at correct size when accessing the Vite dev server directly (`localhost:5173`), but appears scaled when embedded via iframe in the Hermes dashboard. The iframe container gets the correct dimensions (1182×381), and the Three.js canvas CSS/attribute sizes match, yet the visual output looks magnified.

### Tried
- Removed `aspect-ratio`/`max-height` constraints from plugin CSS
- Set canvas CSS size explicitly (`setSize(w, h, true)`)
- Added JS `useEffect` to compute and set iframe container height from viewport position
- None of these resolved the visual scaling in the iframe; the direct Vite page always looks correct

### Suspected
- Something in the Hermes dashboard's page layout or CSS cascade affects the iframe's rendering despite standard viewport/sizing
- Possibly related to device pixel ratio handling, CSS `zoom`, or `font-size`-based scaling on the dashboard page
- Needs further investigation with `contentDocument` inspection or cross-origin iframe debugging

zebster-cmd:
# PR Diagnosis

Verified each review finding against the PR branch. Summary below.

## Verification of reviewer findings

| # | Severity | Location | Status | Notes |
|---|---|---|---|---|
| 1 | 🔴 blocker | `theia-panel/src/ui/FilterBar.ts:133-140` | **CONFIRMED** | The `for (const child of content.children)` loop in `applyBarStyle()` rewrites every label's `cssText` to a fixed string lacking the per-kind `color: state.has(kind) ? fg : fg2`. Initial labels are created at `theia-panel/src/ui/FilterBar.ts:165-170` with the correct color, then `applyBarStyle()` runs at `theia-panel/src/ui/FilterBar.ts:271` (and on every dropdown toggle / theme change / search-visibility change) and strips it. Net effect: unselected kinds appear in default `fg` color, only restored by hover handlers. Real visual regression. |
| 2 | ⚠️ | `theia-panel/src/index.ts:544-547` | **partially incorrect** | The reviewer hedged: "Unless `sidePanel.hide()` invokes the `onClose` callback…". It does. `theia-panel/src/ui/SidePanel.ts:225-233` calls `callbacks?.onClose?.()`, which at `theia-panel/src/index.ts:146-151` calls `searchBar.setPanelOpen(false)` and `filterBar.setSearchToggleVisible(false)` synchronously, un-hiding the input before `.focus()`. Code works, but the implicit ordering dependency is fragile. |
| 3 | ⚠️ | `theia-panel/src/index.ts` | **CONFIRMED** | The pair `searchBar.setPanelOpen(true) + filterBar.setSearchToggleVisible(true)` is duplicated at lines `141-142`, `377-378`, `498-499`, `618-619`. Cleanup is centralized only in the `onClose` handler — asymmetric. Should be wrapped in a helper (e.g. `enterPanelMode()`). |
| 4 | ⚠️ | `theia-panel/src/index.ts:35-50` | **CONFIRMED** | `loadFilterState()` has no schema validation. `new Set(parsed.kinds ?? DEFAULT_KINDS)` accepts an array of arbitrary strings (silently typed as `TheiaGraph["edges"][number]["kind"]` via the function's return type). Also `parsed?.kinds` would be safer if `parsed` were `null`. |
| 5 | ⚠️ | `theia-panel/src/ui/FilterBar.ts:69, 214-216` | **partial** | `separator` is assigned `null` and read in `if (separator)` checks, so noUnusedLocals will *not* trip — but it never receives a non-null value, so all its code paths are no-ops. Vestigial; should be removed. |
| 6 | ⚠️ | `theia-panel/src/index.ts:91-97, 530-548` | **CONFIRMED** | Persisted `kinds` flow into the closure variable on mount (lines 93-97) but the FilterBar UI hard-codes 5 valid kinds at `theia-panel/src/ui/FilterBar.ts:143-149`. A persisted unknown kind survives in the picker/visibility closure but isn't reflected in any toggle, creating silent UI/state divergence until the user changes a filter. |
| 7 | 💡 | `theia-panel/src/ui/SearchBar.ts:231-235` | noted | `wrapper.style.display = panelOpen ? 'none' : ''` relies on `applyWrapperStyle()` (re-run when re-shown) restoring `display: flex`. Works today; brittle to refactors. |
| 8 | 💡 | `theia-panel/src/ui/FilterBar.ts:204-209` | noted | Global `click` listener with `bar.contains()` always-on. HMR / multi-mount could leak; switch to `pointerdown` registered only while `dropdownOpen`. |
| 9 | 💡 | `theia-panel/src/ui/SidePanel.ts:93` | UX | 220ms → 100ms transitions globally; below the typical "deliberate motion" threshold for a panel of this width. |
| 10 | 💡 | `theia-panel/src/index.ts:219-220` | **partial** | `updateVisibility()` clears `focusFilter` unconditionally. Today only two callers exist (`theia-panel/src/index.ts:155` focus-toggle-off, and `:538` filter change with explicit re-apply). Currently safe, but a future caller (e.g. graph reload) would silently kill focus mode. |

## Additional issues observed (not in the review)

- **Unresolved iframe scaling (~125%) in dashboard** (per the author's own comment). Direct Vite access is correct. Several commits attempted fixes (`renderer.setSize(w, h, true)` in `0a474ac`; JS-driven container height in `plugin/src/index.js:221-236` in `7fc13d8`) without resolution. Worth instrumenting `window.devicePixelRatio` and `getComputedStyle(document.documentElement).fontSize` inside the iframe — the symptoms are consistent with a CSS `zoom` / root `font-size` scaling on the dashboard page being inherited through the iframe.
- **`plugin/api/plugin_api.py:32-35`**: mutating global `sys.path` at import is a long-term smell — pollutes search path for whichever process loads the plugin. Cleaner option: `importlib.util.spec_from_file_location` for the sibling `graph_data` module, or restructure as a package. The commit message already flags this as a workaround.
- **`plugin/src/index.js:225-228`**: `resizeContainer` only fires on `window.resize`, not on layout shifts above the iframe (sibling expand/collapse, scrollbar appearance). `ResizeObserver` on the parent would be more robust.
- **Scope creep**: 4–6 of 10 commits (`2c2c2e0`, `7501063`, `142aa51`, `0a474ac`, `7fc13d8`, `7ea36a6`) are dev-environment / iframe-scaling fixes unrelated to the stated #27 / #57 scope. Worth splitting if the iframe-scale issue needs more iteration.

## Recommendation

Block merge until finding #1 is fixed (real regression in the primary filter affordance) and findings #3, #4, #6 are addressed (latent silent bugs). Findings #2, #5, #7–10 are reasonable follow-ups.

Suggested minimal patch for #1: tag each label with `dataset.kind` and re-apply `color: state.has(kind) ? fg : fg2` inside the loop, or stop overwriting `cssText` and set only the affected layout properties via `style.display` etc.

Issue #74, opened 2026-04-25
Focus: mouse-down states invalidated when mouse exits and reenters screen


arm64be:
## Description

Mouse-down interaction states (e.g., dragging nodes, selecting, panning) are not properly validated when the mouse cursor exits the browser window or viewport and then reenters. This can leave the panel in a stuck intermediate state where the UI believes a mouse button is still held down when it is not.

## Expected behavior

When a mouse-down event occurs and then the cursor leaves the screen boundary (e.g., dragging outside the iframe/window) and reenters, the panel should detect this and reset any mid-interaction state. The mouse-down state should be validated on reentry.

## Actual behavior

The panel can remain in a "stuck" state where it believes the mouse button is still pressed, causing erratic behavior on subsequent mouse movements.

## Suggested fix

Listen for `pointerup` / `mouseup` events at the `window` or `document` level (not just on the canvas element) so that button releases outside the viewport are properly handled. Alternatively, validate the `buttons` property on `pointermove` / `mousemove` events to detect when all buttons have been released.

## Environment
- Plugin: theia-constellation

PR #75, opened 2026-04-26
fix(#74): validate mouse-down states when mouse exits and reenters screen

arm64be:
## Summary
- Adds a `resetDrag()` helper function that resets `isMouseDown`, `hasDragged`, and `dragMode`
- Adds a `window`-level `mouseup` listener that calls `resetDrag()` to catch mouse releases outside the element/viewport
- Cleans up the window listener in `destroy()` to prevent leaks

## Details
The panel could enter a stuck drag state when the user mousedown on the canvas, dragged the cursor outside the viewport, and released the button — the `mouseup` event on the element would never fire, leaving `isMouseDown = true` permanently.

The window-level listener ensures mouseup is always caught regardless of cursor position.

### DIFF
--- a/theia-panel/src/index.ts
+++ b/theia-panel/src/index.ts
@@ -470,6 +470,12 @@ export async function mount(
   let mouseDownPos = { x: 0, y: 0 };
   let dragMode: "rotate" | "pan" | null = null;
 
+  function resetDrag() {
+    isMouseDown = false;
+    hasDragged = false;
+    dragMode = null;
+  }
+
   element.addEventListener("mousedown", (e) => {
     if ((e.target as HTMLElement).closest("[data-ui-overlay]")) return;
     if (e.button === 0) dragMode = "rotate";
@@ -500,9 +506,7 @@ export async function mount(
   element.addEventListener("mouseup", (e) => {
     const target = e.target as HTMLElement;
     if (target.closest("aside") || target.closest("[data-ui-overlay]")) {
-      isMouseDown = false;
-      hasDragged = false;
-      dragMode = null;
+      resetDrag();
       return;
     }
     if (isMouseDown && !hasDragged && performance.now() - lastWheelAt >= 200) {
@@ -521,10 +525,11 @@ export async function mount(
         sidePanel.hide();
       }
     }
-    isMouseDown = false;
-    hasDragged = false;
-    dragMode = null;
+    resetDrag();
   });
+
+  // Catch mouseup outside the element/viewport to prevent stuck drag states
+  window.addEventListener("mouseup", resetDrag);
 
   element.addEventListener("contextmenu", (e) => {
     e.preventDefault();
@@ -585,6 +590,7 @@ export async function mount(
   return {
     destroy() {
       disposed = true;
+      window.removeEventListener("mouseup", resetDrag);
       stopThemeListener();
       simulation.stop();
       nodes.dispose();


theia-panel/src/index.ts
```diff
@@ -521,11 +525,12 @@ export async function mount(
...
       }
     }
-    isMouseDown = false;
-    hasDragged = false;
-    dragMode = null;
+    resetDrag();
   });
 
+  // Catch mouseup outside the element/viewport to prevent stuck drag states
+  window.addEventListener("mouseup", resetDrag);
```
zebster-cmd, line 532: **WARNING**: Window-level mouseup only catches releases that happen inside the browser window/document. Issue #74 specifically describes the mouse exiting the *screen/viewport* and re-entering — if the user releases outside the browser entirely, no mouseup fires on window either, and the panel will still be stuck on re-entry. The issue's suggested fix explicitly calls out validating `e.buttons` on pointermove/mousemove as the complementary mechanism for this scenario.

💡 In the existing element-level mousemove handler at line 489, add an early reset when no button is pressed: `if (isMouseDown && e.buttons === 0) { resetDrag(); return; }`. This catches the re-entry case where the release happened outside the browser window.

zebster-cmd, line 532: **SUGGESTION**: Consider using `pointerup` instead of (or alongside) `mouseup`. Pointer events cover mouse, touch, and pen uniformly, and `setPointerCapture` would let the element receive all subsequent pointer events from a drag automatically — eliminating the stuck-state class of bug entirely rather than patching individual scenarios.

💡 window.addEventListener("pointerup", resetDrag); — or refactor mousedown to call element.setPointerCapture(e.pointerId) so the element keeps receiving events even when the cursor leaves it.

theia-panel/src/index.ts
```diff
@@ -521,11 +525,12 @@ export async function mount(
         sidePanel.hide();
       }
     }
-    isMouseDown = false;
-    hasDragged = false;
-    dragMode = null;
+    resetDrag();
```
zebster-cmd, line 528: **SUGGESTION**: On a normal in-element mouseup, both the element-level handler and the window-level handler fire (event bubbles through the document to window), so `resetDrag()` runs twice per release. It's idempotent so this is harmless, but a brief comment on line 532 noting the intentional duplication would help future readers — otherwise it looks like a bug waiting to be 'fixed' by removing one of them.

💡 Add a short comment on line 531 like `// Also fires for in-element mouseups via bubbling — resetDrag is idempotent.`

zebster-cmd (commented):
## Code Review

The refactor extracting `resetDrag()` and adding a window-level `mouseup` listener with proper cleanup is clean and correct as far as it goes. However, this only partially addresses issue #74: if the user releases the mouse button entirely outside the browser window, no `mouseup` fires on `window` either, and the drag state remains stuck on re-entry. The issue itself recommends also validating `e.buttons` on `mousemove`/`pointermove` to catch this case, which is not implemented here.

zebster-cmd:
## Code Review Summary

**Verdict:** COMMENT

The refactor extracting `resetDrag()` and adding a window-level `mouseup` listener with proper cleanup is clean and correct as far as it goes. However, this only partially addresses issue #74: if the user releases the mouse button entirely outside the browser window, no `mouseup` fires on `window` either, and the drag state remains stuck on re-entry. The issue itself recommends also validating `e.buttons` on `mousemove`/`pointermove` to catch this case, which is not implemented here.

### Findings
- ⚠️ **theia-panel/src/index.ts:532** — Window-level mouseup only catches releases that happen inside the browser window/document. Issue #74 specifically describes the mouse exiting the *screen/viewport* and re-entering — if the user releases outside the browser entirely, no mouseup fires on window either, and the panel will still be stuck on re-entry. The issue's suggested fix explicitly calls out validating `e.buttons` on pointermove/mousemove as the complementary mechanism for this scenario.
- 💡 **theia-panel/src/index.ts:532** — Consider using `pointerup` instead of (or alongside) `mouseup`. Pointer events cover mouse, touch, and pen uniformly, and `setPointerCapture` would let the element receive all subsequent pointer events from a drag automatically — eliminating the stuck-state class of bug entirely rather than patching individual scenarios.
- 💡 **theia-panel/src/index.ts:528** — On a normal in-element mouseup, both the element-level handler and the window-level handler fire (event bubbles through the document to window), so `resetDrag()` runs twice per release. It's idempotent so this is harmless, but a brief comment on line 532 noting the intentional duplication would help future readers — otherwise it looks like a bug waiting to be 'fixed' by removing one of them.

Issue #73, opened 2026-04-25
UI inconsistencies with Hermes dashboard


arm64be:
## Description

Several UI elements in the Theia constellation panel don't match the Hermes dashboard's look and feel.

### 1. Font mismatch
The font used in the panel doesn't match the Hermes dashboard font. Should inherit or match the dashboard's font-family.

### 2. Styling colors — yellow accent
Some UI elements use a yellow accent color that doesn't match the Hermes dashboard theme. The visual palette should be consistent with the host dashboard.

### 3. "Focus on connected nodes" placement
The "Focus on connected nodes" toggle/button is currently placed outside of the Filters section. It should be moved into the Filters area for logical grouping.

### 4. Missing iframe border
The iframe that hosts the panel is missing a Hermes dashboard-style border, making it blend poorly with the surrounding dashboard chrome.

## Environment
- Plugin: theia-constellation
- Host: Hermes dashboard

PR #76, opened 2026-04-26
fix(#73): resolve UI inconsistencies with Hermes dashboard

arm64be:
## Summary
Four changes to align the Theia constellation panel with the Hermes dashboard look and feel:

### 1. Font mismatch
Added `Inter` to the default `FONT_STACK` so the panel matches the Hermes dashboard's primary font even without explicit theme forwarding via URL params.

### 2. Yellow accent color
Changed the accent extraction to prefer `--color-primary` and `--color-accent` CSS custom properties before falling back to `--color-warning`. This ensures the panel's accent color matches the dashboard's actual primary/accent color instead of always using the warning yellow.

### 3. Focus on connected nodes placement
Moved the "Focus on connected nodes" toggle from the SidePanel (where it was between the metadata divider and the Connections section) into the FilterBar dropdown (the "Filters" section) for logical grouping with the edge kind toggles. The focus toggle resets when the side panel closes, consistent with the previous behavior.

### 4. Missing iframe border
Added `border` and `border-radius` to `.theia-container` using the dashboard's `--color-border` and `--radius` CSS custom properties, matching the Hermes dashboard card border style.

### DIFF
--- a/plugin/src/style.css
+++ b/plugin/src/style.css
@@ -8,6 +8,8 @@
   overflow: clip;
   box-sizing: border-box;
   background: var(--background-base, #07080d);
+  border: 1px solid var(--color-border, rgba(255, 255, 255, 0.1));
+  border-radius: var(--radius, 0);
 }
 
 .theia-iframe {
@@ -51,12 +53,21 @@
   justify-content: space-between;
   padding: 0.5rem 1rem;
   background: var(--background-base, #07080d);
-  background: color-mix(in srgb, var(--background-base, #07080d) 95%, transparent); /* Chrome 111+, FF 113+, Safari 16.2+ */
+  background: color-mix(
+    in srgb,
+    var(--background-base, #07080d) 95%,
+    transparent
+  ); /* Chrome 111+, FF 113+, Safari 16.2+ */
   border-bottom: 1px solid var(--color-border, rgba(255, 255, 255, 0.1));
   color: var(--midground-base, #cfd6e4);
-  font-family: system-ui, -apple-system, sans-serif;
+  font-family:
+    system-ui,
+    -apple-system,
+    sans-serif;
   z-index: 1;
-  transition: background 200ms, border-color 200ms;
+  transition:
+    background 200ms,
+    border-color 200ms;
 }
 
 .theia-iframe-full {
--- a/theia-panel/src/index.ts
+++ b/theia-panel/src/index.ts
@@ -177,9 +177,7 @@ export async function mount(
     },
     onClose: () => {
       if (focusEnabled) {
-        focusEnabled = false;
-        focusFilter = null;
-        updateVisibility();
+        onFocusToggle(false);
         filterBar.setFocusEnabled(false);
       }
       clearSelected();
@@ -565,13 +563,15 @@ export async function mount(
       if (id) applyFocusModeIfEnabled(id);
     },
     theme,
-    modelFilter,
-    () => {
-      sidePanel.hide();
-      searchBar.input.focus();
+    {
+      initialModel: modelFilter,
+      onSearchToggle: () => {
+        sidePanel.hide();
+        searchBar.input.focus();
+      },
+      onFocusToggle,
+      initialFocusEnabled: focusEnabled,
     },
-    onFocusToggle,
-    focusEnabled,
   );
 
   const listeners: Record<string, Array<(...args: unknown[]) => void>> = {
--- a/theia-panel/src/index.ts
+++ b/theia-panel/src/index.ts
@@ -148,6 +148,17 @@ export async function mount(
     filterBar.setSearchToggleVisible(true);
   }
 
+  function onFocusToggle(enabled: boolean) {
+    focusEnabled = enabled;
+    if (!focusEnabled) {
+      focusFilter = null;
+      updateVisibility();
+    } else {
+      const id = sidePanel.currentNodeId();
+      if (id) applyFocusModeIfEnabled(id);
+    }
+  }
+
   const sidePanel = createSidePanel(element, theme, {
     onNavigate: (targetId) => {
       const idx = nodeIndex.get(targetId);
@@ -165,20 +176,16 @@ export async function mount(
       emit("node-click", targetId);
     },
     onClose: () => {
+      if (focusEnabled) {
+        focusEnabled = false;
+        focusFilter = null;
+        updateVisibility();
+        filterBar.setFocusEnabled(false);
+      }
       clearSelected();
       searchBar.setPanelOpen(false);
       filterBar.setSearchToggleVisible(false);
       nodes.flush();
-    },
-    onFocusToggle: (enabled) => {
-      focusEnabled = enabled;
-      if (!focusEnabled) {
-        focusFilter = null;
-        updateVisibility();
-      } else {
-        const id = sidePanel.currentNodeId();
-        if (id) applyFocusModeIfEnabled(id);
-      }
     },
   });
 
@@ -563,6 +570,8 @@ export async function mount(
       sidePanel.hide();
       searchBar.input.focus();
     },
+    onFocusToggle,
+    focusEnabled,
   );
 
   const listeners: Record<string, Array<(...args: unknown[]) => void>> = {
--- a/theia-panel/src/ui/FilterBar.ts
+++ b/theia-panel/src/ui/FilterBar.ts
@@ -66,7 +66,8 @@ export function createFilterBar(
   initialTheme: ThemeTokens,
   options: FilterBarOptions = {},
 ) {
-  const { initialModel, onSearchToggle, onFocusToggle, initialFocusEnabled } = options;
+  const { initialModel, onSearchToggle, onFocusToggle, initialFocusEnabled } =
+    options;
   let theme = initialTheme;
   const bar = document.createElement("div");
   bar.dataset.uiOverlay = "";
@@ -301,9 +302,7 @@ export function createFilterBar(
       background: ${focusEnabled ? `#${theme.accent}` : `#${theme.bg}`};
       cursor: pointer; transition: background .15s, border-color .15s;
     `;
-    focusToggleEl.style.color = focusEnabled
-      ? `#${theme.fg}`
-      : `#${theme.fg2}`;
+    focusToggleEl.style.color = focusEnabled ? `#${theme.fg}` : `#${theme.fg2}`;
   }
 
   function initFocusToggle() {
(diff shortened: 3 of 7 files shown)


theia-panel/src/ui/FilterBar.ts
```diff
@@ -290,17 +332,30 @@ export function createFilterBar(
...
   }
 
+  function setFocusEnabled(enabled: boolean) {
+    focusEnabled = enabled;
+    if (focusToggleEl) {
+      const cb = focusToggleEl.querySelector(
+        "input[type=checkbox]",
+      ) as HTMLInputElement;
+      if (cb) cb.checked = enabled;
+      focusToggleEl.style.color = enabled ? `#${theme.fg}` : `#${theme.fg2}`;
```
zebster-cmd: **CRITICAL**: setFocusEnabled syncs cb.checked and label color but not cb.style.background. The user-driven onchange handler sets background to theme.accent / theme.bg, so when focus is externally disabled (e.g., from index.ts onClose -> filterBar.setFocusEnabled(false)), the checkbox stays visually filled with the accent color even though it's unchecked. This produces a desynced UI on every panel close while focus mode is on.

💡 Inside setFocusEnabled, after setting cb.checked, also update the background:
  if (cb) {
    cb.checked = enabled;
    cb.style.background = enabled ? `#${theme.accent}` : `#${theme.bg}`;
  }

theia-panel/src/ui/FilterBar.ts
```diff
@@ -290,17 +332,30 @@ export function createFilterBar(
...
+    }
+  }
+
   function updateTheme(newTheme: ThemeTokens) {
     theme = newTheme;
     applyBarStyle();
     for (const t of toggles) t._applyStyle();
     applySelectStyle();
+    rebuildFocusToggle();
```
zebster-cmd: **WARNING**: updateTheme calls rebuildFocusToggle(), which removes and re-creates the <label>+<input> on every theme mutation. Other elements in this file (toggles, model select) just restyle in place via _applyStyle / applySelectStyle. The MutationObserver in plugin/src/index.js fires on :root style/class changes and forwards tokens, so theme updates can be frequent — rebuilding tears down the checkbox listener and triggers extra layout each time.

💡 Mirror the pattern used by toggles: extract an applyFocusToggleStyle() that re-reads `theme` and rewrites cssText + cb.style on the existing focusToggleEl, and call that from updateTheme instead of rebuildFocusToggle().

theia-panel/src/index.ts
```diff
@@ -563,6 +570,8 @@ export async function mount(
       sidePanel.hide();
       searchBar.input.focus();
     },
+    onFocusToggle,
+    focusEnabled,
```
zebster-cmd: **WARNING**: createFilterBar now has 9 positional parameters, with onSearchToggle, onFocusToggle, and initialFocusEnabled all trailing optionals of similar shape. A future caller that passes onFocusToggle without onSearchToggle (or swaps the boolean position) will silently bind to the wrong slot — TypeScript can't catch that since the types are compatible across positions.

💡 Consider collapsing the optional tail into an options bag, e.g. createFilterBar(container, models, onModelChange, onSearchOpen, theme, { initialModel, onSearchToggle, onFocusToggle, initialFocusEnabled }). Can be done as a follow-up, but worth flagging now that this signature is growing.

theia-panel/src/index.ts
```diff
@@ -165,21 +176,17 @@ export async function mount(
       emit("node-click", targetId);
     },
     onClose: () => {
+      if (focusEnabled) {
+        focusEnabled = false;
+        focusFilter = null;
```
zebster-cmd: **WARNING**: onClose now handles focus disable, but it doesn't guard against re-entry via filterBar.setFocusEnabled(false) → cb.onchange. Verify that programmatically setting cb.checked inside setFocusEnabled does not fire the onchange handler (it shouldn't — the `change` event is only dispatched on user input — but if the implementation later switches to `cb.click()` or dispatches an event, this would loop through onClose again).

💡 No code change required today; just keep setFocusEnabled to property assignment (cb.checked = …) and avoid dispatching synthetic events on it.

theia-panel/src/ui/Theme.ts
```diff
@@ -29,7 +29,7 @@ export interface ThemeTokens {
 }
 
 export const FONT_STACK =
-  "system-ui, -apple-system, 'Segoe UI', Roboto, 'Helvetica Neue', Arial, sans-serif";
+  "Inter, system-ui, -apple-system, 'Segoe UI', Roboto, 'Helvetica Neue', Arial, sans-serif";
```
zebster-cmd, line 32: **SUGGESTION**: FONT_STACK now leads with `Inter` but there is no @font-face / @import in this package and the panel runs in a sandboxed iframe (sandbox="allow-scripts allow-same-origin" per plugin/src/index.js). If the host dashboard hasn't loaded Inter into the iframe context, this prepend is a silent no-op and the panel falls back to system-ui — fine, but it means the visual change won't actually land in the panel UI without an accompanying font load.

💡 Either ship an @font-face for Inter in theia-panel's CSS, document the dashboard-side dependency, or drop the prepend to keep the stack honest.

plugin/src/style.css
```diff
@@ -8,6 +8,8 @@
   overflow: clip;
   box-sizing: border-box;
   background: var(--background-base, #07080d);
+  border: 1px solid var(--color-border, rgba(255, 255, 255, 0.1));
+  border-radius: var(--radius, 0);
```
zebster-cmd, line 12: **SUGGESTION**: Adding `border: 1px solid var(--color-border, …)` and `border-radius: var(--radius, 0)` directly on .theia-container may double-up if the iframe content already paints its own outer chrome (the panel uses ThemeTokens.border for its own surfaces). Worth a visual check at both radius=0 and radius>0 dashboard themes.

💡 If the iframe panel already renders an outer border, drop the container border and rely on the inner one (or vice versa) to avoid a 2px nested look.

theia-panel/src/ui/FilterBar.ts
```diff
@@ -274,10 +277,49 @@ export function createFilterBar(
...
+      focusEnabled = cb.checked;
+      cb.style.background = focusEnabled ? `#${theme.accent}` : `#${theme.bg}`;
+      focusToggleEl!.style.color = focusEnabled
+        ? `#${theme.fg}`
+        : `#${theme.fg2}`;
+      onFocusToggle?.(focusEnabled);
+    };
+    focusToggleEl.append(
+      cb,
+      document.createTextNode("Focus on connected nodes"),
```
zebster-cmd, line 320: **SUGGESTION**: The focus toggle label uses `font: 10px/1.4 var(--theia-font, ${FONT_STACK})` while the rest of the bar's font sizing comes from applyBarStyle(). Hard-coding 10px here means a future bar-wide font-size change won't carry to this row. Minor consistency nit.

💡 Consider inheriting font-size or using the same scale token already established in applyBarStyle.

zebster-cmd (dismissed):
## Code Review

The PR migrates the focus-mode toggle from SidePanel into FilterBar and adds dashboard-aligned chrome (border, Inter font, accent fallbacks). The refactor is structurally sound, but `FilterBar.setFocusEnabled` does not sync the checkbox's background fill, so when the side panel closes with focus active, the checkbox visually remains "on" while the state is off. Also, `createFilterBar` now takes 9 positional parameters and `rebuildFocusToggle` runs on every theme tick — both worth tightening before merge.

zebster-cmd:
## Code Review Summary

**Verdict:** REQUEST_CHANGES

The PR migrates the focus-mode toggle from SidePanel into FilterBar and adds dashboard-aligned chrome (border, Inter font, accent fallbacks). The refactor is structurally sound, but `FilterBar.setFocusEnabled` does not sync the checkbox's background fill, so when the side panel closes with focus active, the checkbox visually remains "on" while the state is off. Also, `createFilterBar` now takes 9 positional parameters and `rebuildFocusToggle` runs on every theme tick — both worth tightening before merge.

### Findings
- 🔴 **theia-panel/src/ui/FilterBar.ts:342** — setFocusEnabled syncs cb.checked and label color but not cb.style.background. The user-driven onchange handler sets background to theme.accent / theme.bg, so when focus is externally disabled (e.g., from index.ts onClose -> filterBar.setFocusEnabled(false)), the checkbox stays visually filled with the accent color even though it's unchecked. This produces a desynced UI on every panel close while focus mode is on.
- ⚠️ **theia-panel/src/ui/FilterBar.ts:351** — updateTheme calls rebuildFocusToggle(), which removes and re-creates the <label>+<input> on every theme mutation. Other elements in this file (toggles, model select) just restyle in place via _applyStyle / applySelectStyle. The MutationObserver in plugin/src/index.js fires on :root style/class changes and forwards tokens, so theme updates can be frequent — rebuilding tears down the checkbox listener and triggers extra layout each time.
- ⚠️ **theia-panel/src/index.ts:574** — createFilterBar now has 9 positional parameters, with onSearchToggle, onFocusToggle, and initialFocusEnabled all trailing optionals of similar shape. A future caller that passes onFocusToggle without onSearchToggle (or swaps the boolean position) will silently bind to the wrong slot — TypeScript can't catch that since the types are compatible across positions.
- ⚠️ **theia-panel/src/index.ts:181** — onClose now handles focus disable, but it doesn't guard against re-entry via filterBar.setFocusEnabled(false) → cb.onchange. Verify that programmatically setting cb.checked inside setFocusEnabled does not fire the onchange handler (it shouldn't — the `change` event is only dispatched on user input — but if the implementation later switches to `cb.click()` or dispatches an event, this would loop through onClose again).
- 💡 **theia-panel/src/ui/Theme.ts:32** — FONT_STACK now leads with `Inter` but there is no @font-face / @import in this package and the panel runs in a sandboxed iframe (sandbox="allow-scripts allow-same-origin" per plugin/src/index.js). If the host dashboard hasn't loaded Inter into the iframe context, this prepend is a silent no-op and the panel falls back to system-ui — fine, but it means the visual change won't actually land in the panel UI without an accompanying font load.
- 💡 **plugin/src/style.css:12** — Adding `border: 1px solid var(--color-border, …)` and `border-radius: var(--radius, 0)` directly on .theia-container may double-up if the iframe content already paints its own outer chrome (the panel uses ThemeTokens.border for its own surfaces). Worth a visual check at both radius=0 and radius>0 dashboard themes.
- 💡 **theia-panel/src/ui/FilterBar.ts:313** — The focus toggle label uses `font: 10px/1.4 var(--theia-font, ${FONT_STACK})` while the rest of the bar's font sizing comes from applyBarStyle(). Hard-coding 10px here means a future bar-wide font-size change won't carry to this row. Minor consistency nit.

PR #79, opened 2026-04-28
feat: add Focus on search toggle to filter bar

arm64be:
## Summary

- Adds a "Focus on search" checkbox in the filter dropdown next to the existing "Focus on connected nodes" checkbox
- When enabled and a search query is active, hides all nodes that don't match the current search
- Reactively updates visibility as the user types in the search bar
- Works alongside the existing "Focus on connected nodes" filter (both can be active simultaneously)

### Changes

- **`theia-panel/src/ui/SearchBar.ts`**: Exposes `getMatchedNodeIds(query)` for reuse
- **`theia-panel/src/ui/FilterBar.ts`**: Adds checkbox UI, state, and `onSearchFocusToggle`/`setSearchFocusEnabled` API
- **`theia-panel/src/index.ts`**: Wires up the toggle, applies search filter in `updateVisibility` and when query changes

### DIFF
--- a/theia-panel/src/index.ts
+++ b/theia-panel/src/index.ts
@@ -110,6 +110,7 @@ export async function mount(
     modelFilter = saved.model;
   }
   let focusEnabled = false;
+  let searchFocusEnabled = false;
   let focusFilter: Set<string> | null = null;
 
   // Mutable graph-specific state — closures capture the binding, not the value
@@ -157,6 +158,13 @@ export async function mount(
       const id = sidePanel.currentNodeId();
       if (id) applyFocusModeIfEnabled(id);
     }
+  }
+
+  function onSearchFocusToggle(enabled: boolean) {
+    searchFocusEnabled = enabled;
+    updateVisibility();
+    const id = sidePanel.currentNodeId();
+    if (id) applyFocusModeIfEnabled(id);
   }
 
   const sidePanel = createSidePanel(element, theme, {
@@ -245,6 +253,17 @@ export async function mount(
 
   function updateVisibility() {
     visibleNodeIds = computeVisibleNodeIds(currentGraph, kinds, modelFilter);
+    if (searchFocusEnabled && searchBar) {
+      const query = searchBar.input.value.trim();
+      if (query) {
+        const matchedIds = searchBar.getMatchedNodeIds(query);
+        const filtered = new Set<string>();
+        for (const id of visibleNodeIds) {
+          if (matchedIds.has(id)) filtered.add(id);
+        }
+        visibleNodeIds = filtered;
+      }
+    }
     for (let i = 0; i < currentGraph.nodes.length; i++) {
       nodes.setVisible(i, visibleNodeIds.has(currentGraph.nodes[i]!.id));
     }
@@ -403,6 +422,13 @@ export async function mount(
       theme,
       (node) => visibleNodeIds.has(node.id),
     );
+    searchBar.input.addEventListener("input", () => {
+      if (searchFocusEnabled) {
+        updateVisibility();
+        const id = sidePanel.currentNodeId();
+        if (id) applyFocusModeIfEnabled(id);
+      }
+    });
   }
 
   function createLoadingOverlay(text: string): { remove(): void } {
@@ -576,6 +602,8 @@ export async function mount(
       },
       onFocusToggle,
       initialFocusEnabled: focusEnabled,
+      onSearchFocusToggle,
+      initialSearchFocusEnabled: searchFocusEnabled,
     },
   );
 
--- a/theia-panel/src/index.ts
+++ b/theia-panel/src/index.ts
@@ -43,6 +43,7 @@ const STORAGE_KEY = "theia-constellation-filter";
 function loadFilterState(): {
   kinds: Set<TheiaGraph["edges"][number]["kind"]>;
   model: string | null;
+  searchFocus: boolean;
 } | null {
   try {
     const raw = localStorage.getItem(STORAGE_KEY);
@@ -57,17 +58,22 @@ function loadFilterState(): {
     return {
       kinds: new Set(parsedKinds.length > 0 ? parsedKinds : DEFAULT_KINDS),
       model: typeof parsed?.model === "string" ? parsed.model : null,
+      searchFocus: parsed?.searchFocus === true,
     };
   } catch {
     return null;
   }
 }
 
-function saveFilterState(kinds: Set<string>, model: string | null): void {
+function saveFilterState(
+  kinds: Set<string>,
+  model: string | null,
+  searchFocus: boolean,
+): void {
   try {
     localStorage.setItem(
       STORAGE_KEY,
-      JSON.stringify({ kinds: Array.from(kinds), model }),
+      JSON.stringify({ kinds: Array.from(kinds), model, searchFocus }),
     );
   } catch {
     /* quota exceeded, ignore */
@@ -110,8 +116,11 @@ export async function mount(
     modelFilter = saved.model;
   }
   let focusEnabled = false;
-  let searchFocusEnabled = false;
+  let searchFocusEnabled = saved?.searchFocus ?? false;
   let focusFilter: Set<string> | null = null;
+  let searchFocusMatchKey = "";
+  let searchInputController: AbortController | null = null;
+  let searchFocusTimer: ReturnType<typeof setTimeout> | null = null;
 
   // Mutable graph-specific state — closures capture the binding, not the value
   let currentGraph: TheiaGraph;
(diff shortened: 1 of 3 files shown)


theia-panel/src/index.ts
```diff
@@ -245,6 +253,17 @@ export async function mount(
...
     visibleNodeIds = computeVisibleNodeIds(currentGraph, kinds, modelFilter);
+    if (searchFocusEnabled && searchBar) {
+      const query = searchBar.input.value.trim();
+      if (query) {
+        const matchedIds = searchBar.getMatchedNodeIds(query);
+        const filtered = new Set<string>();
+        for (const id of visibleNodeIds) {
+          if (matchedIds.has(id)) filtered.add(id);
+        }
+        visibleNodeIds = filtered;
```
zebster-cmd: **WARNING**: When searchFocusEnabled is true and the query has no matches, `visibleNodeIds` is replaced with an empty Set, hiding every node in the graph. A user typing a single non-matching character will see the graph blank out, which is confusing — there is no visual cue that the disappearance is due to a filter mismatch vs. data loss.

💡 Skip the assignment when matchedIds is empty (`if (matchedIds.size === 0) { /* keep original visibility */ } else { visibleNodeIds = filtered; }`), or surface a 'no matches' indicator in the SearchBar so the empty state is explicit.

theia-panel/src/index.ts
```diff
@@ -403,6 +422,13 @@ export async function mount(
       theme,
       (node) => visibleNodeIds.has(node.id),
     );
+    searchBar.input.addEventListener("input", () => {
```
zebster-cmd: **WARNING**: The input listener fires on every keystroke and triggers `updateVisibility()` → `getMatchedNodeIds()` (O(N) over all graph.nodes) plus `applyFocusModeIfEnabled()`. For large constellations this can cause typing lag and unnecessary WebGL state churn while the user is mid-word.

💡 Debounce the handler (~100–150ms via setTimeout/clearTimeout or rAF coalescing), and clear the pending timer in the search bar's dispose path.

theia-panel/src/index.ts
```diff
@@ -403,6 +422,13 @@ export async function mount(
...
+      if (searchFocusEnabled) {
+        updateVisibility();
+        const id = sidePanel.currentNodeId();
+        if (id) applyFocusModeIfEnabled(id);
```
zebster-cmd: **WARNING**: `applyFocusModeIfEnabled(id)` is invoked on every keystroke when searchFocusEnabled is on. If focusEnabled is also active, this re-computes the focus neighborhood for the current node on each character — independent of whether the search result actually changes the focus subject.

💡 Only call applyFocusModeIfEnabled when the matched-set or current node identity actually changes (compare prior matchedIds size/membership), or move it outside the per-keystroke path.

zebster-cmd (commented):
## Code Review

The "Focus on search" feature is wired through cleanly and follows the existing focus-toggle patterns in FilterBar/SearchBar. The main concerns are UX around empty match results hiding the entire graph, per-keystroke O(N) cost on the search input, and a missing cleanup path for the new input listener. No correctness blockers, but a couple of warnings worth addressing before merge.

theia-panel/src/index.ts
zebster-cmd: **SUGGESTION**: The newly added `addEventListener('input', ...)` has no matching `removeEventListener`. If `initSearchBar` (the enclosing function) is ever re-invoked without disposing the prior `searchBar.input`, listeners accumulate. Even if today it's a single-shot init, an explicit teardown keeps this resilient.

💡 Use an AbortController whose signal is passed to addEventListener, and call `controller.abort()` from the search bar's dispose path — or expose the handler so it can be removed alongside `searchBar.dispose()`.

theia-panel/src/index.ts
```diff
@@ -110,6 +110,7 @@ export async function mount(
     modelFilter = saved.model;
   }
   let focusEnabled = false;
+  let searchFocusEnabled = false;
```
zebster-cmd: **SUGGESTION**: `searchFocusEnabled` is initialized to false locally but never read from or written to `saved` state, while sibling filter state (`modelFilter`) does persist. If the intent is for toggle preferences to survive reloads, this is an inconsistency; if not, ignore.

💡 If persistence is desired, restore from `saved.searchFocus` on mount and write it back inside `onSearchFocusToggle` alongside the existing save logic.

theia-panel/src/ui/SearchBar.ts
```diff
@@ -246,5 +246,17 @@ export function createSearchBar(
     container.removeChild(wrapper);
   }
 
-  return { updateTheme, setPanelOpen, dispose, input };
+  function getMatchedNodeIds(query: string): Set<string> {
```
zebster-cmd, line 258: **SUGGESTION**: `getMatchedNodeIds` re-iterates `graph.nodes` on every call. The SearchBar already runs `matches(node, query)` when rendering its own results list, so the same work is done twice per keystroke.

💡 Cache the most recent matched-id Set when the SearchBar computes its own results, and have `getMatchedNodeIds` return the cached value (recomputing only on cache miss / different query).

theia-panel/src/ui/FilterBar.ts
```diff
@@ -305,6 +316,25 @@ export function createFilterBar(
...
+      letter-spacing: 0.05em;
+      color: ${searchFocusEnabled ? `#${theme.fg}` : `#${theme.fg2}`};
+    `;
+    searchFocusCb.style.cssText = `
+      appearance: none; width: 14px; height: 14px; margin: 0; flex-shrink: 0;
+      border: 1px solid #${theme.border};
+      border-radius: ${theme.radius};
+      background: ${searchFocusEnabled ? `#${theme.accent}` : `#${theme.bg}`};
+      cursor: pointer; transition: background .15s, border-color .15s;
+    `;
```
zebster-cmd: **SUGGESTION**: `applySearchFocusToggleStyle` rewrites the full `cssText` on every theme update, which negates the `transition: background .15s, border-color .15s` because the element's previous styles are wiped before transition can interpolate. This mirrors the existing focus toggle so it's not a regression, but worth noting.

💡 Set only the changing properties individually (`searchFocusCb.style.background = ...; searchFocusCb.style.borderColor = ...;`) so transitions actually animate, and keep the static layout rules in a one-time setup pass.

theia-panel/src/ui/FilterBar.ts
```diff
@@ -323,11 +353,30 @@ export function createFilterBar(
...
+    applySearchFocusToggleStyle();
+    content.append(searchFocusToggleEl);
+  }
+
   applyBarStyle();
   dropdown.appendChild(content);
   bar.append(btn, searchToggle, dropdown);
   rebuildModelSelect();
   initFocusToggle();
+  initSearchFocusToggle();
```
zebster-cmd: **SUGGESTION**: `initSearchFocusToggle()` is always called regardless of whether `onSearchFocusToggle` was provided. The toggle then renders but its onchange becomes a no-op, which can mislead users.

💡 Gate `initSearchFocusToggle()` on `onSearchFocusToggle != null`, matching how a feature flag would be expressed elsewhere.

zebster-cmd:
## Code Review Summary

**Verdict:** COMMENT

The "Focus on search" feature is wired through cleanly and follows the existing focus-toggle patterns in FilterBar/SearchBar. The main concerns are UX around empty match results hiding the entire graph, per-keystroke O(N) cost on the search input, and a missing cleanup path for the new input listener. No correctness blockers, but a couple of warnings worth addressing before merge.

### Findings
- ⚠️ **theia-panel/src/index.ts:264** — When searchFocusEnabled is true and the query has no matches, `visibleNodeIds` is replaced with an empty Set, hiding every node in the graph. A user typing a single non-matching character will see the graph blank out, which is confusing — there is no visual cue that the disappearance is due to a filter mismatch vs. data loss.
- ⚠️ **theia-panel/src/index.ts:425** — The input listener fires on every keystroke and triggers `updateVisibility()` → `getMatchedNodeIds()` (O(N) over all graph.nodes) plus `applyFocusModeIfEnabled()`. For large constellations this can cause typing lag and unnecessary WebGL state churn while the user is mid-word.
- ⚠️ **theia-panel/src/index.ts:429** — `applyFocusModeIfEnabled(id)` is invoked on every keystroke when searchFocusEnabled is on. If focusEnabled is also active, this re-computes the focus neighborhood for the current node on each character — independent of whether the search result actually changes the focus subject.
- 💡 **theia-panel/src/index.ts:425** — The newly added `addEventListener('input', ...)` has no matching `removeEventListener`. If `initSearchBar` (the enclosing function) is ever re-invoked without disposing the prior `searchBar.input`, listeners accumulate. Even if today it's a single-shot init, an explicit teardown keeps this resilient.
- 💡 **theia-panel/src/index.ts:113** — `searchFocusEnabled` is initialized to false locally but never read from or written to `saved` state, while sibling filter state (`modelFilter`) does persist. If the intent is for toggle preferences to survive reloads, this is an inconsistency; if not, ignore.
- 💡 **theia-panel/src/ui/SearchBar.ts:249** — `getMatchedNodeIds` re-iterates `graph.nodes` on every call. The SearchBar already runs `matches(node, query)` when rendering its own results list, so the same work is done twice per keystroke.
- 💡 **theia-panel/src/ui/FilterBar.ts:332** — `applySearchFocusToggleStyle` rewrites the full `cssText` on every theme update, which negates the `transition: background .15s, border-color .15s` because the element's previous styles are wiped before transition can interpolate. This mirrors the existing focus toggle so it's not a regression, but worth noting.
- 💡 **theia-panel/src/ui/FilterBar.ts:379** — `initSearchFocusToggle()` is always called regardless of whether `onSearchFocusToggle` was provided. The toggle then renders but its onchange becomes a no-op, which can mislead users.

PR #82, opened 2026-05-02
[codex] Animate onboarding edge links

arm64be:
## Summary
- add per-edge reveal progress to the Three.js edge layer
- animate onboarding connections so links draw in after both endpoint nodes are revealed
- reset edge reveal state after onboarding completes

## Validation
- npm run format:check
- npm run typecheck

### DIFF
--- a/theia-panel/src/index.ts
+++ b/theia-panel/src/index.ts
@@ -41,6 +41,7 @@ const ONBOARDING_STORAGE_KEY = "theia-first-load-onboarding-complete";
 const PHYSICS_SNAPSHOT_KEY_PREFIX = "theia-physics-snapshot:";
 const ONBOARDING_ROTATION_RADIANS = Math.PI * 1.15;
 const ONBOARDING_BLINK_MS = 3200;
+const ONBOARDING_LINK_UP_MS = 700;
 const PHYSICS_SNAPSHOT_INTERVAL_MS = 5000;
 const ONBOARDING_BASE_ZOOM = 0.72;
 const ONBOARDING_MIN_ZOOM = 0.42;
@@ -63,6 +64,12 @@ function revealBrightness(t: number): number {
   if (t >= 1) return 1;
   const eased = easeQuadInOut(t);
   return 1 + 0.5 * Math.sin(Math.PI * eased);
+}
+
+function edgeKey(edge: TheiaGraph["edges"][number]): string {
+  return edge.source < edge.target
+    ? `${edge.source}|${edge.target}|${edge.kind}`
+    : `${edge.target}|${edge.source}|${edge.kind}`;
 }
 
 function hasCompletedOnboarding(): boolean {
@@ -185,6 +192,7 @@ export async function mount(
     rankByIndex: Map<number, number>;
     revealedNodeIds: Set<string>;
     revealStartedAtByIndex: Map<number, number>;
+    linkStartedAtByKey: Map<string, number>;
     lastRevealedCount: number;
     lastEase: number;
     cameraZoom: number;
@@ -624,6 +632,7 @@ export async function mount(
       rankByIndex: new Map(order.map((index, rank) => [index, rank])),
       revealedNodeIds: new Set(),
       revealStartedAtByIndex: new Map(),
+      linkStartedAtByKey: new Map(),
       lastRevealedCount: 0,
       lastEase: 0,
       cameraZoom: ONBOARDING_BASE_ZOOM,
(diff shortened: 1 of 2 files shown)


theia-panel/src/index.ts
```diff
@@ -65,6 +66,12 @@ function revealBrightness(t: number): number {
   return 1 + 0.5 * Math.sin(Math.PI * eased);
 }
 
+function edgeKey(edge: TheiaGraph["edges"][number]): string {
```
zebster-cmd, line 70: **Hot-path string allocation.** During onboarding, `edgeKey()` is invoked ~3×E times per frame: once per edge in the loop at L658, once per edge inside the `setConnectionProgress` callback (called from `Edges.ts` L246), and once per edge from `updatePositions` (`Edges.ts` L191). At the 1000+ edge target that's ~3000 transient strings/frame.

Cheap fix: precompute the key once per edge in `beginOnboarding` (e.g. `Map<GraphEdge, string>`) or stash it on the edge object.

theia-panel/src/index.ts
```diff
@@ -641,6 +650,31 @@ export async function mount(
     replaceSimulation(activeVisibleNodeIds(), true);
   }
 
+  function updateOnboardingLinks(now: number) {
+    if (!onboarding) {
+      edges.setConnectionProgress(null);
```
zebster-cmd, line 661: Minor: `setConnectionProgress(null)` here, then again at L759 when `onboarding = null`. Idempotent, but `setConnectionProgress` walks every line × every valid index even when called with `null` (`Edges.ts` L228+). Could short-circuit when the callback didn't actually change.

theia-panel/src/index.ts
```diff
@@ -641,6 +650,31 @@ export async function mount(
...
+      return;
+    }
+    for (const edge of currentGraph.edges) {
+      if (
+        onboarding.revealedNodeIds.has(edge.source) &&
+        onboarding.revealedNodeIds.has(edge.target)
```
zebster-cmd, line 667: Nit: this `revealedNodeIds.has(source) && has(target)` check is already implicit because `rebuildVisibleEdges` only produces geometry for edges in `activeVisibleNodeIds()` (which is intersected with `revealedNodeIds`). Dead defense; safe to keep for clarity, but worth a comment so a future reader doesn't think it's load-bearing.

theia-panel/src/index.ts
```diff
@@ -701,6 +736,7 @@ export async function mount(
       onboarding.overlay.remove();
       setNodeVisibilityFromState();
       rebuildVisibleEdges();
+      updateOnboardingLinks(now);
```
zebster-cmd, line 745: **Possible visual pulse at the completion boundary.** When `raw >= 1`, the block above marks every remaining node revealed and then this call stamps every still-unstarted edge with `linkStartedAtByKey.set(key, now)` — so any edge whose endpoints were not yet both revealed begins a fresh 700 ms fade-in **after** the overlay says 100%.

The post-completion blink loop keeps the frame loop alive long enough that this likely works out fine, but if the intent (per the PR body) is *reset* edge reveal state, consider stamping these edges with `now - ONBOARDING_LINK_UP_MS` so they snap to full instead of starting a new animation.

theia-panel/src/index.ts
```diff
@@ -720,6 +756,7 @@ export async function mount(
           nodes.setBrightness(idx, 1);
         }
         onboarding = null;
+        edges.setConnectionProgress(null);
```
zebster-cmd, line 765: See L655 — second of two `setConnectionProgress(null)` calls. If you decide to short-circuit when the callback doesn't change, this one becomes a no-op too.

theia-panel/src/scene/Edges.ts
```diff
@@ -173,8 +186,17 @@ export function createEdges(): EdgeLayer {
...
+          Math.min(1, getConnectionProgress?.(e) ?? 1),
+        );
         posAttr.setXYZ(i * 2 + 0, sx, sy, sz);
-        posAttr.setXYZ(i * 2 + 1, tx, ty, tz);
+        posAttr.setXYZ(
+          i * 2 + 1,
+          sx + (tx - sx) * progress,
+          sy + (ty - sy) * progress,
+          sz + (tz - sz) * progress,
+        );
```
zebster-cmd, line 199: Question: was animating both the line endpoint position **and** the alpha (via `aReveal`) intentional? Lines grow out from the source while fading in — nice effect, but it doubles the work and could be done with alpha alone. If both are wanted, no objection, but worth a one-line comment so the duplication doesn't read as accidental.

theia-panel/src/scene/Edges.ts
```diff
@@ -205,6 +227,31 @@ export function createEdges(): EdgeLayer {
     }
   }
```
zebster-cmd, line 228: Naming nit: `setConnectionProgress` reads ambiguously alongside `updatePositions` / `setHoverNode`. `setEdgeRevealProgress` would map more cleanly to the `aReveal` attribute it drives. Take or leave.

zebster-cmd (commented):
Inline review (posted on behalf of @zebster-cmd by Claude). Net take: ship-ready — no blockers, the integration matches existing patterns nicely. Notes are tightening passes.

One out-of-diff note that doesn't fit on a line: `edgeKey()` keys by `(source, target, kind)`, but the dedup in `Edges.ts` (~L91–94) keys by `(source, target)` only. This is fine in practice because `rebuild` filters by kind first, so the surviving edge always carries the matching kind — but a one-line comment noting that invariant would save the next reader 5 minutes.

PR #83, opened 2026-05-02
[codex] Persist selected session focus

arm64be:
## Summary

- make the selected node keep the same edge-focus dimming behavior used by hover after the pointer leaves
- make the side-panel Go to session control reuse the same camera-focus navigation path as connection-list session clicks

## Root Cause

The selection path only marked the node as selected and did not keep the edge hover focus active after hover cleared. The Go to session control emitted the host node-click event directly instead of going through the panel navigation routine that selects and focuses the node.

## Validation

- npm run typecheck
- npm run format:check
- npm test
- npm run build

### DIFF
--- a/theia-panel/src/index.ts
+++ b/theia-panel/src/index.ts
@@ -406,7 +406,10 @@ export async function mount(
   }
 
   function select(idx: number) {
-    clearSelected();
+    if (selectedIdx !== null && selectedIdx !== idx) {
+      nodes.setSelected(selectedIdx, false);
+      nodes.setHighlight(selectedIdx, false);
+    }
     selectedIdx = idx;
     nodes.setSelected(idx, true);
     nodes.setHighlight(idx, false);
@@ -1173,8 +1176,7 @@ export async function mount(
       if (selectedId !== null) {
         const idx = nodeIndex.get(selectedId);
         if (idx !== undefined) {
-          selectedIdx = idx;
-          nodes.setSelected(idx, true);
+          select(idx);
           const n = currentGraph.nodes[idx]!;
           const related = currentGraph.edges.filter(
             (e) =>
--- a/theia-panel/src/index.ts
+++ b/theia-panel/src/index.ts
@@ -391,12 +391,18 @@ export async function mount(
   }
 
   const tooltip = createTooltip(element, theme);
+
+  function selectedNodeId(): string | null {
+    return selectedIdx === null ? null : currentGraph.nodes[selectedIdx]!.id;
+  }
+
   function clearSelected() {
     if (selectedIdx !== null) {
       nodes.setSelected(selectedIdx, false);
       nodes.setHighlight(selectedIdx, false);
       selectedIdx = null;
     }
+    edges.setHoverNode(null);
   }
 
   function select(idx: number) {
@@ -404,6 +410,7 @@ export async function mount(
     selectedIdx = idx;
     nodes.setSelected(idx, true);
     nodes.setHighlight(idx, false);
+    edges.setHoverNode(currentGraph.nodes[idx]!.id);
   }
 
   function enterPanelMode(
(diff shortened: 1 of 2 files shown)


theia-panel/src/index.ts
```diff
@@ -391,19 +391,26 @@ export async function mount(
...
     }
+    edges.setHoverNode(null);
   }
 
   function select(idx: number) {
     clearSelected();
     selectedIdx = idx;
     nodes.setSelected(idx, true);
     nodes.setHighlight(idx, false);
+    edges.setHoverNode(currentGraph.nodes[idx]!.id);
```
zebster-cmd, line 416: Minor perf nit: `clearSelected()` at L409 now sets hover to `null` (L405), and immediately after we set it back to the new id here. `setHoverNode` walks every edge of every kind (`Edges.ts` L182–205), so this doubles that work on every selection.

Fix: inline a single `setHoverNode(currentGraph.nodes[idx]!.id)` after the prior-selection cleanup, or skip the clear step's edge update when the caller is `select()`.

Related: see the summary — the reload restore path at L1172–L1184 doesn't go through `select()`, so it misses the new invariant entirely. Replacing that block with `select(idx)` would dedupe both concerns.

theia-panel/src/index.ts
```diff
@@ -434,21 +441,28 @@ export async function mount(
     if (id && searchMatchesChanged) applyFocusModeIfEnabled(id);
   }
 
+  function navigateToNode(targetId: string) {
+    const idx = nodeIndex.get(targetId);
+    if (idx === undefined || !activeVisibleNodeIds().has(targetId)) return;
```
zebster-cmd, line 449: **"Go to session" silently no-ops when the selected node is filtered out.** If the user enables a filter that hides the currently-selected node, the SidePanel still renders (selection persists) and the button is shown — but `navigateToNode` bails here because `activeVisibleNodeIds()` doesn't contain `targetId`.

Either disable the button when `!activeVisibleNodeIds().has(currentId)` in `SidePanel.ts`, or have the filter callback (around L1083–L1108) clear/hide the selection when it's no longer visible.

theia-panel/src/index.ts
```diff
@@ -804,7 +819,7 @@ export async function mount(
     picker.onHover((idx) => {
       const nodeId = idx === null ? null : currentGraph.nodes[idx]!.id;
       element.style.cursor = idx === null ? "" : "pointer";
-      edges.setHoverNode(nodeId);
+      edges.setHoverNode(nodeId ?? selectedNodeId());
```
zebster-cmd, line 825: Question: this hover-leave correctly falls back to `selectedNodeId()`, which is great — but the search-driven `enterPanelMode` path doesn't go through `select(idx)`, so `selectedIdx` stays `null` and the new edge-dim invariant never kicks in for search-opened panels.

Is that intentional asymmetry, or should search also `select(idx)`?

theia-panel/src/ui/SidePanel.ts
```diff
@@ -132,6 +133,10 @@ export function createSidePanel(
     }
   }
 
+  function handleGoToSession() {
+    if (currentId) callbacks?.onGoToSession?.(currentId);
```
zebster-cmd, line 137: Question: why mirror navigation as a separate `onGoToSession(nodeId)` callback when it always points at `currentId` and just re-runs `navigateToNode`?

A no-arg callback (or just `onGoToSession()` reading `panel.currentNodeId()` upstream) would let `SidePanel` not depend on knowing the `currentId` semantics. Open to leaving as-is for symmetry with `onNavigate`, but worth a sentence in the PR body if so.

zebster-cmd (commented):
Inline review (posted on behalf of @zebster-cmd by Claude). One real bug worth fixing before merge — see below — plus polish items as inline threads.

**Bug — reload restore path skips the new edge-dim invariant.** The PR establishes a new invariant: whenever a node is selected, `edges.setHoverNode(selectedNodeId())` should be active (see `select()` at L413, `clearSelected()` at L405, `rebuildVisibleEdges()` at L594, picker hover at L822). But the reload re-selection block (`index.ts` L1172–L1184, unchanged code outside this diff) re-establishes `selectedIdx` and calls `nodes.setSelected(idx, true)` directly — never calls `edges.setHoverNode(...)`. After a reload that re-selects, edges will be undimmed, the opposite of the new invariant.

Cleanest fix: replace the manual setup in that block with a call to `select(idx)`, which already does both. That also dedupes the reselect logic.

---

**Meta nits:**
- PR title says *Persist selected session focus*, but no persistence (localStorage/etc.) is added — the change is in-memory edge-dim continuity + a Go-to-session button. Body is accurate; title is the part that's misleading.
- This branch's parent (`21383a8`) is several commits ahead of `main`'s `33364eb`, so reviewing via `gh pr diff` alone is misleading — pull the branch.

PR #99, opened 2026-05-03
fix(panel): linear group reveal rate — fix empty start + bursty middle

zebster-cmd:
## Summary
The cubic group rate from PR #98 produced the exact two symptoms the user reported:

- **Empty first phase**: `easeInOutCubic(0.1) ≈ 0.004` — only ~6 of 1600 nodes are revealed in the first 10% of duration. The user sees nothing happening for the first ~2 seconds, then a flood.
- **Bursty middle**: peak rate is `1.5 × N / duration` around `t=0.5` (~109 nodes/sec at 1.6k/22s). This concentrated arrival rate fires `replaceActive` rebuilds rapid-succession on the worker and stacks dozens of concurrent per-node animations — the FPS-tanking "burst."

Linear group rate gives a uniform `N / duration` reveal rate from the first frame onward (~73 nodes/sec at 1.6k/22s, ~1.2 per frame at 60Hz). No empty start, no concentrated mid-flood.

Per-node `popScale` (easeQuadInOut over 600ms wall-clock) is unchanged — it still smooths each individual node's visual entry.

Removed the now-unused `easeInOutCubic` helper and updated stale comments referring to the supernova/cubic curves.

## Test plan
- [ ] Nodes start spawning immediately at t=0 — no dead first phase.
- [ ] Spawn rate stays roughly constant throughout — no perceptible mid-onboarding flood.
- [ ] FPS stays stable across the full reveal (no dips at the previous "burst" moment).
- [ ] Each individual node still eases in smoothly over ~600ms.
- [ ] Total duration unchanged (5-22s based on node count); overlay percent climbs evenly.

### DIFF
--- a/theia-panel/src/state/onboarding.ts
+++ b/theia-panel/src/state/onboarding.ts
@@ -11,8 +11,8 @@ const ONBOARDING_BLINK_MS = 3200;
 const ONBOARDING_LINK_UP_MS = 700;
 // Camera zoom progressively retreats across the reveal so the user can
 // watch the constellation expand into view. Driven by raw time progress
-// (not reveal fraction) so the zoom-out feels steady regardless of how
-// the supernova-burst-then-easeInOutCubic reveal lands nodes.
+// (not reveal fraction) so the zoom-out feels steady regardless of the
+// reveal cadence.
 const ONBOARDING_START_ZOOM = 0.72;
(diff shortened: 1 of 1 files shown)


zebster-cmd:
Linear was still too fast at start — replacing with quadratic (t²) curve in a follow-up.

PR #100, opened 2026-05-03
fix(panel): quadratic group reveal rate for gradual ramp-up

zebster-cmd:
## Summary
Replaces #99 (closed). Linear was still too dense at the very start — at 1.6k nodes / 22s, the first 100ms revealed ~7 nodes which read as "much at once". `easeInOutCubic` (PR #98) was the opposite extreme: `cubic(0.1) ≈ 0.004` left the first ~2s essentially empty before flooding the worker mid-reveal.

Quadratic `t²` sits between them and gives the gradual ramp-up the user described ("1, then 2, then 4…"):

| time | revealed (of 1.6k) |
|------|--------------------|
| 1.1s |   4 |
| 2.2s |  16 |
| 4.4s |  64 |
| 11s  | 400 |
| 22s  | 1600 |

End-of-reveal rate is ~2× linear (last second adds ~140 nodes), but per-node `easeQuadInOut` over 600ms wall-clock keeps each individual entry visually smooth even at the higher group rate.

Removed the now-unused `easeInOutCubic` helper and updated stale comments.

## Test plan
- [ ] First 1-2 seconds: see ~1-4 nodes appear individually (not a wall of nodes).
- [ ] Spawn rate gradually accelerates throughout — no perceived "burst" mid-onboarding.
- [ ] Per-node entries still smooth (~600ms ease).
- [ ] FPS stable across the full reveal.
- [ ] Total duration unchanged (5-22s based on node count).

### DIFF
--- a/theia-panel/src/state/onboarding.ts
+++ b/theia-panel/src/state/onboarding.ts
@@ -11,8 +11,8 @@ const ONBOARDING_BLINK_MS = 3200;
 const ONBOARDING_LINK_UP_MS = 700;
 // Camera zoom progressively retreats across the reveal so the user can
 // watch the constellation expand into view. Driven by raw time progress
-// (not reveal fraction) so the zoom-out feels steady regardless of how
-// the supernova-burst-then-easeInOutCubic reveal lands nodes.
+// (not reveal fraction) so the zoom-out feels steady regardless of the
+// reveal cadence.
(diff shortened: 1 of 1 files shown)


zebster-cmd:
Consolidating with the per-node fast-pop + 24s duration changes in a follow-up PR.

